PR #42, opened 2026-03-19
Fix/macdive uddf import (Fixes #28)

### LINKED ISSUE #28: Dive profile is empty after importing UDDF exported from MacDive.
Dive profile is empty after importing UDDF exported from MacDive.

https://scubaboard.com/community/threads/submersion-free-open-source-dive-log-app-all-platforms-looking-for-dive-computer-testers.667061/post-10730136

steinbil:
## Summary

Adds a normalizer step before the UDDF parser, to normalize different flavors of UDDF from other dive logging software, and implements the specific MacDive normalizer to fix the issue of Macdive imports not showing dive profile, bottom time or runtime. Ready to add support for more flavors of UDDF in the future if necessary.

## Changes

- Add UddfNormalizer dispatcher that detects the UDDF dialect and routes to the appropriate handler before parsing
- Add MacDiveDialectNormalizer that pre-processes MacDive UDDF exports:
- Strips the default XML namespace (xmlns="http://www.streit.cc/uddf/3.2/") that caused all findElements() calls to silently return empty results
- Moves <country> from geography/address/country to a direct child of <site> where the parser expects it
- Copies <equipmentused> from informationafterdive into informationbeforedive where the parser reads it
- Fix <diveduration> parsing to handle float strings (e.g. "3494.00") — previously int.tryParse() returned null, leaving runtime empty
- Fix <divetime> waypoint parsing to handle float strings — previously all profile timestamps collapsed to 0, breaking the dive profile graph
- Add 25 unit tests covering dialect detection, each normalizer transformation, and end-to-end import of depth, duration, temperature, notes, site, buddy, gas mix, weight, and profile waypoints

## Test Plan

- [x] `flutter test` passes
- [x] Manual testing on: iPhone 16e, iOS 26.3 simulator

Fixes #28

### DIFF
--- a/lib/core/services/export/uddf/dialects/macdive_dialect.dart
+++ b/lib/core/services/export/uddf/dialects/macdive_dialect.dart
@@ -31,7 +31,43 @@ class MacDiveDialect extends UddfDialect {
 
   @override
   bool isMatch(XmlDocument doc) {
-    return doc.rootElement.namespaceUri == _namespace;
+    if (doc.rootElement.namespaceUri != _namespace) return false;
+
+    // UDDF-compliant exporters include <generator><name> identifying the app
+    // (e.g. Subsurface, divelog_convert). When present, use it as the
+    // authoritative signal to avoid false positives on other apps' exports.
+    final generators = doc.findAllElements('generator');
+    if (generators.isNotEmpty) {
+      return generators.any((gen) {
+        final name = gen.findElements('name').firstOrNull?.innerText ?? '';
+        return name.toLowerCase().contains('macdive');
+      });
+    }
+
+    // No <generator> tag: fall back to structural quirk detection for files
+    // that lack one (e.g. older MacDive versions or hand-edited UDDF).
+    return _hasMacDiveStructuralQuirks(doc);
+  }
+
+  // Returns true when the document exhibits structural deviations unique to
+  // MacDive: country nested under geography/address instead of directly under
+  // site, or equipmentused placed in informationafterdive.
+  bool _hasMacDiveStructuralQuirks(XmlDocument doc) {
+    final hasNestedCountry = doc.findAllElements('site').any((site) {
+      final geo = site.findElements('geography').firstOrNull;
+      if (geo == null) return false;
+      final address = geo.findElements('address').firstOrNull;
+      if (address == null) return false;
+      return address.findElements('country').isNotEmpty;
+    });
+
+    final hasEquipmentInAfter = doc.findAllElements('dive').any((dive) {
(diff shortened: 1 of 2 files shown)


ericgriffin:
Regarding missing data/fields, some apps imply values rather than stating them.

For example:
- Subsurface UDDF may omit surfacetemperature but include it in waypoint data, so a dialect-aware handler could infer it from the first waypoint
- Some apps omit altitude entirely, implicitly meaning sea-level diving (0m), while others may omit gas mix, implying air (21/0)
- MacDive may encode duration differently than Diving Log, so "missing" fields might actually be present but in a non-standard location

The distinction between "empty" vs "missing" is also a real concern in XML. A <depth></depth> tag vs no <depth> tag at all could need different handling depending on the dialect.

steinbil:
After some back and forth with Claude, we tried a different approach with the same goal. It handles structural, encoding and missing-fields categories as part of the normalization step, using dialect classes. This keeps the parser completely unaware of any dialects — it simply handles standard UDDF.

**As I write this I realize I haven't really implemented this in the way you suggested - by keeping everything in the pre-parser stage, it can't inject "domain-level defaults" or lean on the Submersion database. However, I'm not sure that's a problem.** The examples you listed could all be handled with only the XML in the pre-parser stage. And the thing I keep returning to is this:

_Who makes a big multi-dive UDDF import? A new user migrating from a different app. They are not going to have anything set up in Submersion that would be relevant for interpreting missing data or injecting any defaults. And as such, all we have to work with is in the XML and can be handled pre-parsing._ 

Please let me know if I'm thinking about this in the wrong way, or if there's still something I'm not seeing. My perspective on this from the start has been UDDF import as a migration tool - since that's the reality that I'm facing.

As for extra data/fields: I haven't found any MacDive-specific non-standard extra fields yet, so nothing is implemented there. If that changes, the current approach could be extended to map extra non-standard fields to Submersion's custom fields feature as part of normalizeXml — a useful fallback to preserve data that has no direct mapping. More structured handling of non-standard fields would likely need to live outside the dialect layer, but I think we can cross that bridge when we get there?

PS.
I am also not a pro when it comes to git, so Claude might have been a bit heavy handed with some force-pushes etc, just to remove a lingering line you added which is no longer needed. Hope it doesn't look too bad.

PPS
Like I said, I'm leaning very heavily on Claude for the actual code, since I'm not familiar with dart. I'm just using what little I know from computer science classes and my own hacky projects to try to steer Claude in the right direction, with a view to following the style of the codebase. If it turns out you don't think this code is up to par, I guess I might have to learn some dart and get more familiar with the codebase before I do any more damage...

PR #170, opened 2026-04-05
Fix/additional uddf ssrf properties

hancockks:
## Summary

Improves imported-profile parity between UDDF and SSRF by wiring several direct sample and dive-level fields into the existing backend model, and updates the superpowers tracker to distinguish real implementation gaps from format/model `N/A` cases.

### UDDF
- Persist sample heart rate
- Parse and persist sample `cns`
- Parse and persist sample `ndl`
- Parse and persist sample `rbt`
- Persist dive-level `cns` / `otu` into the source snapshot
- Map `decostop@kind` to `decoType`
  - `safetystop -> 1`
  - any other present `decostop` kind -> `2`
  - no `decostop` leaves `decoType` as `null`
- Log a warning for unsupported UDDF `decostop@kind` values while still preserving the broader deco semantic

### SSRF / Subsurface XML
- Map `divecomputer@dctype` to `diveMode`
- Parse and persist dive-level `cns` / `otu`
- Parse and persist sample `ndl`
- Parse and persist sample `tts`
- Parse and persist sample `rbt`
- Parse and persist sample `cns`
- Parse and persist sample heart rate from `heartbeat`
- Map cylinder `use` to tank role
- Map `in_deco=1` to `decoType=2`
  - `in_deco=0` remains `null`
- Map sample `po2` to `ppO2`

### Documentation
- Added and updated the import-gap tracker in `docs/superpowers/specs/2026-04-05-imported-profile-gap-priority-tracker.md`
- Clarified `Yes / No / N/A` support semantics
- Split `ceiling` from `next stop`
- Marked non-actionable rows as blank in the `Fixed` column
- Documented the UDDF `rbt` heuristic:
  - prefer `remainingbottomtime`
  - fall back to `remainingo2time`

## Why

The backend already supported far more imported-profile fidelity than UDDF/SSRF were using. This PR closes the low-risk, direct field-mapping gaps first, especially around deco-state, oxygen-toxicity, and CCR-related sample data.

## Verification

```bash
flutter test test/core/services/export/uddf/uddf_full_import_service_test.dart test/features/dive_import/data/services/uddf_entity_importer_test.dart
flutter test test/features/universal_import/data/parsers/subsurface_xml_parser_test.dart
dart analyze lib/core/services/export/uddf/uddf_full_import_service.dart lib/features/dive_import/data/services/uddf_entity_importer.dart test/core/services/export/uddf/uddf_full_import_service_test.dart test/features/dive_import/data/services/uddf_entity_importer_test.dart
dart analyze lib/features/universal_import/data/parsers/subsurface_xml_parser.dart test/features/universal_import/data/parsers/subsurface_xml_parser_test.dart
```

## Follow-Ups

- SSRF rebreather semantics beyond direct `ppO2`
- UDDF/SSRF `next stop` support if we add a backend field
- Dive-level deco metadata / provenance completeness
- Profile events / markers persistence

## Screen shots
Before:
image

After: (cns now shown -- shows need for not mixing calc & DC)
image

### DIFF
--- a/lib/core/utils/number_utils.dart
+++ b/lib/core/utils/number_utils.dart
@@ -1,0 +1,8 @@
+double? asDoubleOrNull(Object? value) {
+  if (value is num) return value.toDouble();
+  return null;
+}
+
+extension NullableNumParsing on Object? {
+  double? toDoubleOrNull() => asDoubleOrNull(this);
+}
--- a/lib/features/dive_import/data/services/uddf_entity_importer.dart
+++ b/lib/features/dive_import/data/services/uddf_entity_importer.dart
@@ -4,6 +4,7 @@ import 'package:submersion/core/database/database.dart'
     show DiveDataSourcesCompanion;
 import 'package:submersion/core/services/export/export_service.dart';
 import 'package:submersion/core/services/location_service.dart';
+import 'package:submersion/core/utils/number_utils.dart';
 import 'package:submersion/features/buddies/data/repositories/buddy_repository.dart';
 import 'package:submersion/features/buddies/domain/entities/buddy.dart';
 import 'package:submersion/features/certifications/data/repositories/certification_repository.dart';
@@ -185,8 +186,6 @@ class UddfEntityImporter {
   final TankPresetEntity? _defaultTankPreset;
   final int _defaultStartPressure;
   final bool _applyDefaultTankToImports;
-
-  static double? _asDouble(Object? value) => (value as num?)?.toDouble();
 
   UddfEntityImporter({
     TankPresetEntity? defaultTankPreset,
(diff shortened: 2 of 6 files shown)


ericgriffin:
For this PR I think we should keep these new imports:
  - heartRate parsing and persistence (both formats) — pure sensor data
  - diveMode mapping from SSRF dctype — metadata
  - Tank role mapping from SSRF cylinder use — metadata
  - The _asDouble safety casting — good defensive fix
  - The _parseDouble percent-stripping fix

But not include these (unless there's some really good reason to import them them):
  - Sample-level cns, ndl, rbt, tts — these should be calculated by Submersion's own engine
  - Dive-level cnsEnd and otu — same reasoning
  - decoType/in_deco — DCs do report this, but so does Subsurface's recalculation

ericgriffin:
> > e DC data in the after screenshot, why wouldn't we interpolate the missing sample points to avoid the (incorrect) down spikes?
> 
> The algorithm (which hasn't change in this PR) is if the toggle is DC to use DC and fill in from computed. Since the above profile has a tank switch, CNS is still computing off the first tank -- the downward spikes are correct, the calculations are incorrect.

gotcha - makes sense

PR #207, opened 2026-04-12
feat: prefer CR5L BLE characteristics on macOS

KalenJosifovski:
## Summary

Prefer the CREST CR5L Nordic UART Service characteristics in the macOS BLE bridge.

This is the Submersion-side companion to pending `libdivecomputer` CR5L support. On its own, this PR does not add CR5L parsing or transport to Submersion. It adjusts the Darwin BLE bridge so that once the vendored `libdivecomputer` is updated to a CR5L-capable revision, Submersion selects the correct CoreBluetooth service/characteristics for the watch.

This PR should not be merged before the corresponding `libdivecomputer` CR5L support is available and vendored into Submersion.

## Changes

- Add CR5L-specific BLE characteristic preferences in the Darwin bridge.
- Prefer the CR5L Nordic UART Service UUIDs when the device is `CREST-CR5L`.
- Subscribe correctly to the CR5L response characteristics so downstream `libdivecomputer` transport can receive control responses, list packets, and bulk data packets.

## Why

The CR5L uses a Nordic UART Service layout with multiple response paths:

- Service: `6E400001-B5A3-F393-E0A9-E50E24DCCA9E`
- Write: `6E400002-B5A3-F393-E0A9-E50E24DCCA9E`
- Response / notify-indicate paths:
  - `6E400002-B5A3-F393-E0A9-E50E24DCCA9E`
  - `6E400003-B5A3-F393-E0A9-E50E24DCCA9E`
  - `6E400004-B5A3-F393-E0A9-E50E24DCCA9E`

Without this preference logic, the macOS bridge can connect to the watch but route packets incorrectly, which prevents the downstream CR5L transport from completing a download.

## Scope

Included in this PR:

- `packages/libdivecomputer_plugin/darwin/Sources/LibDCDarwin/BleIoStream.swift`

Explicitly not included in this PR:

- local development override files such as `LIBDIVECOMPUTER_FORK_PATH.txt`
- local `libdivecomputer` fork build wiring used only for development
- vendored `third_party/libdivecomputer` updates
- unrelated macOS build workaround changes

## Dependency

This PR depends on the corresponding `libdivecomputer` CR5L support PR. During local development, Submersion was tested against an external local `libdivecomputer` fork via an untracked pointer file and build override. That local development mechanism is intentionally not part of this PR.

The intended merge sequence is:

1. Merge `libdivecomputer` CR5L support upstream.
2. Vendor/update that `libdivecomputer` revision into Submersion.
3. Merge this Submersion BLE bridge change alongside the vendor update if still needed.

## Background Material

Optional background material for reviewers who want the reverse-engineering and validation context:

- [Full reverse-engineering report](https://gist.github.com/de364160b2942b6bd7ec509b8e216dea)
- [Protocol note](https://gist.github.com/fcf1a9c2288bdaf457a9e504ac4900bf)
- [File-format note](https://gist.github.com/c4bb80ec26b964793a9eeedfd5040aa1)
- [Live validation note](https://gist.github.com/0ca9f6213be02bfb9bf10e3f5e2df256)

These are intentionally not required to review this PR. The code change here is narrow and only handles the macOS BLE bridge side.

## Test Plan

- [x] `flutter test` passes
- [ ] `flutter analyze` passes (failed only because of the unrelated local-only `packages/google_sign_in_ios_patched
        package`)
- [x] Manual macOS testing against a real `CREST-CR5L` watch using a local CR5L-enabled `libdivecomputer` fork

Manual testing performed locally:

- Submersion discovered `CREST-CR5L`
- macOS CoreBluetooth selected the intended CR5L NUS characteristics
- end-to-end download succeeded downstream once paired with the local CR5L-enabled `libdivecomputer`

## Notes For Reviewers

- This PR is intentionally narrow.
- The actual CR5L protocol implementation lives in the companion `libdivecomputer` [PR](https://github.com/libdivecomputer/libdivecomputer/pull/59).
- If reviewers prefer, this PR can also wait until the Submersion vendored `libdivecomputer` subtree is updated in a follow-up PR.

### DIFF
--- a/packages/libdivecomputer_plugin/darwin/Sources/LibDCDarwin/BleIoStream.swift
+++ b/packages/libdivecomputer_plugin/darwin/Sources/LibDCDarwin/BleIoStream.swift
@@ -16,18 +16,24 @@ class BleIoStream: NSObject, CBPeripheralDelegate {
 
     private static let preferredServiceUUIDs: Set<CBUUID> = [
         CBUUID(string: "CB3C4555-D670-4670-BC20-B61DBC851E9A"),
+        CBUUID(string: "6E400001-B5A3-F393-E0A9-E50E24DCCA9E"),
     ]
 
     private static let preferredWriteUUIDs: Set<CBUUID> = [
         CBUUID(string: "6606AB42-89D5-4A00-A8CE-4EB5E1414EE0"),
+        CBUUID(string: "6E400002-B5A3-F393-E0A9-E50E24DCCA9E"),
     ]
 
     private static let preferredNotifyUUIDs: Set<CBUUID> = [
         CBUUID(string: "A60B8E5C-B267-44D7-9764-837CAF96489E"),
+        CBUUID(string: "6E400003-B5A3-F393-E0A9-E50E24DCCA9E"),
     ]
 
     private static let pelagicGen1TxCharacteristic = CBUUID(
         string: "6606AB42-89D5-4A00-A8CE-4EB5E1414EE0"
+    )
+    private static let crestCr5lTxCharacteristic = CBUUID(
+        string: "6E400002-B5A3-F393-E0A9-E50E24DCCA9E"
     )
     private static let notifySettleDelaySeconds: TimeInterval = 0.3
     private static let bleIoctlType: UInt32 = UInt32(Character("b").asciiValue!)
@@ -45,12 +51,14 @@ class BleIoStream: NSObject, CBPeripheralDelegate {
 
     private var writeCharacteristic: CBCharacteristic?
     private var notifyCharacteristic: CBCharacteristic?
+    private var subscribedCharacteristics: [CBCharacteristic] = []
     private var bestCandidate: CharacteristicCandidate?
     private var remainingServiceDiscoveries = 0
     private var discoverySignaled = false
 
     private let readLock = NSLock()
-    private var readBuffer = Data()
+    private var readPackets: [Data] = []
+    private var partialReadBuffer = Data()
     private let readSemaphore = DispatchSemaphore(value: 0)
     private let writeSemaphore = DispatchSemaphore(value: 0)
     private let writeReadySemaphore = DispatchSemaphore(value: 0)
@@ -138,6 +146,7 @@ class BleIoStream: NSObject, CBPeripheralDelegate {
         bestCandidate = nil
         writeCharacteristic = nil
         notifyCharacteristic = nil
+        subscribedCharacteristics = []
         hasSeenNotify = false
         writeWithoutResponsePreferred = false
         consecutiveReadTimeouts = 0
@@ -261,9 +270,9 @@ class BleIoStream: NSObject, CBPeripheralDelegate {
     private static let cPoll: libdc_io_poll_fn = { userdata, timeout in
         let stream = Unmanaged<BleIoStream>.fromOpaque(userdata!).takeUnretainedValue()
         stream.readLock.lock()
-        let available = stream.readBuffer.count
+        let available = !stream.partialReadBuffer.isEmpty || !stream.readPackets.isEmpty
         stream.readLock.unlock()
-        if available > 0 { return Int32(LIBDC_STATUS_SUCCESS) }
+        if available { return Int32(LIBDC_STATUS_SUCCESS) }
 
         if timeout == 0 { return Int32(LIBDC_STATUS_TIMEOUT) }
 
@@ -286,13 +295,15 @@ class BleIoStream: NSObject, CBPeripheralDelegate {
 
         while true {
             readLock.lock()
-            let available = readBuffer.count
-            if available > 0 {
-                let bytesToRead = min(size, readBuffer.count)
-                _ = readBuffer.withUnsafeBytes { ptr in
+            if partialReadBuffer.isEmpty, !readPackets.isEmpty {
+                partialReadBuffer = readPackets.removeFirst()
+            }
+            if !partialReadBuffer.isEmpty {
+                let bytesToRead = min(size, partialReadBuffer.count)
+                _ = partialReadBuffer.withUnsafeBytes { ptr in
                     memcpy(data, ptr.baseAddress!, bytesToRead)
                 }
-                readBuffer.removeFirst(bytesToRead)
+                partialReadBuffer.removeFirst(bytesToRead)
                 readLock.unlock()
                 actual.pointee = bytesToRead
                 return Int32(LIBDC_STATUS_SUCCESS)
@@ -392,7 +403,8 @@ class BleIoStream: NSObject, CBPeripheralDelegate {
         }
 
         readLock.lock()
-        readBuffer.removeAll(keepingCapacity: true)
+        readPackets.removeAll(keepingCapacity: true)
+        partialReadBuffer.removeAll(keepingCapacity: true)
         readLock.unlock()
         drainSemaphore(readSemaphore)
         return Int32(LIBDC_STATUS_SUCCESS)
@@ -431,6 +443,9 @@ class BleIoStream: NSObject, CBPeripheralDelegate {
 
     private func initialWriteWithoutResponsePreference(for characteristic: CBCharacteristic) -> Bool {
         let properties = characteristic.properties
+        if characteristic.uuid == Self.crestCr5lTxCharacteristic {
+            return false
+        }
         if characteristic.uuid == Self.pelagicGen1TxCharacteristic {
             return true
         }
@@ -626,7 +641,7 @@ class BleIoStream: NSObject, CBPeripheralDelegate {
                     + " \(error.localizedDescription)")
             return
         }
-        if let notify = notifyCharacteristic, characteristic.uuid != notify.uuid {
+        guard subscribedCharacteristics.contains(where: { $0.uuid == characteristic.uuid }) else {
             return
         }
         guard let value = characteristic.value else { return }
@@ -637,7 +652,7 @@ class BleIoStream: NSObject, CBPeripheralDelegate {
                 + " bytes=\(value.count)"
                 + " data=\(Self.hexString(value, maxBytes: Self.maxLogBytes))")
         readLock.lock()
-        readBuffer.append(value)
+        readPackets.append(value)
         readLock.unlock()
         readSemaphore.signal()
     }
@@ -660,11 +675,10 @@ class BleIoStream: NSObject, CBPeripheralDelegate {
     func peripheral(_ peripheral: CBPeripheral,
                     didUpdateNotificationStateFor characteristic: CBCharacteristic,
                     error: Error?) {
-        guard let notify = notifyCharacteristic, characteristic.uuid == notify.uuid else {
+        guard subscribedCharacteristics.contains(where: { $0.uuid == characteristic.uuid }) else {
             return
         }
         guard waitingForNotifyEnable else { return }
-        waitingForNotifyEnable = false
 
         if let error {
             NativeLogger.e("BleIoStream", category: "BLE",
@@ -681,8 +695,11 @@ class BleIoStream: NSObject, CBPeripheralDelegate {
         }
         NativeLogger.d("BleIoStream", category: "BLE",
             "Notify enabled for \(characteristic.uuid.uuidString)")
-        isReady = true
-        signalDiscoveryReady()
+        if subscribedCharacteristics.allSatisfy(\.isNotifying) {
+            waitingForNotifyEnable = false
+            isReady = true
+            signalDiscoveryReady()
+        }
     }
 
     private func signalDiscoveryReady() {
@@ -749,6 +766,17 @@ class BleIoStream: NSObject, CBPeripheralDelegate {
 
         writeCharacteristic = candidate.write
         notifyCharacteristic = candidate.notify
+        let serviceCharacteristics = candidate.write.service?.characteristics ?? []
+        if candidate.write.uuid == Self.crestCr5lTxCharacteristic {
+            subscribedCharacteristics = serviceCharacteristics.filter {
+                $0.properties.contains(.notify) || $0.properties.contains(.indicate)
+            }
+            if subscribedCharacteristics.isEmpty {
+                subscribedCharacteristics = [candidate.notify]
+            }
+        } else {
+            subscribedCharacteristics = [candidate.notify]
+        }
         writeWithoutResponsePreferred = initialWriteWithoutResponsePreference(for: candidate.write)
         NativeLogger.d("BleIoStream", category: "BLE",
             "Selected write=\(candidate.write.uuid.uuidString)"
@@ -757,15 +785,20 @@ class BleIoStream: NSObject, CBPeripheralDelegate {
                 + " (\(Self.propertySummary(candidate.notify.properties)))"
                 + " (score=\(candidate.score))")
         NativeLogger.d("BleIoStream", category: "BLE",
+            "Subscribed characteristics:"
+                + " \(subscribedCharacteristics.map { $0.uuid.uuidString }.joined(separator: ","))")
+        NativeLogger.d("BleIoStream", category: "BLE",
             "Initial write mode preference:"
                 + " \(writeWithoutResponsePreferred ? "withoutResponse" : "withResponse")")
-        if candidate.notify.isNotifying {
+        if !subscribedCharacteristics.isEmpty && subscribedCharacteristics.allSatisfy(\.isNotifying) {
             isReady = true
             signalDiscoveryReady()
             return
         }
         waitingForNotifyEnable = true
-        peripheral.setNotifyValue(true, for: candidate.notify)
+        for characteristic in subscribedCharacteristics where !characteristic.isNotifying {
+            peripheral.setNotifyValue(true, for: characteristic)
+        }
     }
 }
 
--- a/packages/libdivecomputer_plugin/darwin/Sources/LibDCDarwin/BleIoStream.swift
+++ b/packages/libdivecomputer_plugin/darwin/Sources/LibDCDarwin/BleIoStream.swift
@@ -16,24 +16,27 @@ class BleIoStream: NSObject, CBPeripheralDelegate {
 
     private static let preferredServiceUUIDs: Set<CBUUID> = [
         CBUUID(string: "CB3C4555-D670-4670-BC20-B61DBC851E9A"),
-        CBUUID(string: "6E400001-B5A3-F393-E0A9-E50E24DCCA9E"),
     ]
 
     private static let preferredWriteUUIDs: Set<CBUUID> = [
         CBUUID(string: "6606AB42-89D5-4A00-A8CE-4EB5E1414EE0"),
-        CBUUID(string: "6E400002-B5A3-F393-E0A9-E50E24DCCA9E"),
     ]
 
     private static let preferredNotifyUUIDs: Set<CBUUID> = [
         CBUUID(string: "A60B8E5C-B267-44D7-9764-837CAF96489E"),
-        CBUUID(string: "6E400003-B5A3-F393-E0A9-E50E24DCCA9E"),
     ]
 
     private static let pelagicGen1TxCharacteristic = CBUUID(
         string: "6606AB42-89D5-4A00-A8CE-4EB5E1414EE0"
     )
-    private static let crestCr5lTxCharacteristic = CBUUID(
+    private static let nordicUartServiceUUID = CBUUID(
+        string: "6E400001-B5A3-F393-E0A9-E50E24DCCA9E"
+    )
+    private static let nordicUartServiceRxCharacteristic = CBUUID(
         string: "6E400002-B5A3-F393-E0A9-E50E24DCCA9E"
+    )
+    private static let nordicUartServiceNotifyCharacteristic = CBUUID(
+        string: "6E400003-B5A3-F393-E0A9-E50E24DCCA9E"
     )
     private static let notifySettleDelaySeconds: TimeInterval = 0.3
     private static let bleIoctlType: UInt32 = UInt32(Character("b").asciiValue!)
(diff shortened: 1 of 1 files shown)


packages/libdivecomputer_plugin/darwin/Sources/LibDCDarwin/BleIoStream.swift
```diff
@@ -16,19 +16,25 @@ class BleIoStream: NSObject, CBPeripheralDelegate {
...
         CBUUID(string: "A60B8E5C-B267-44D7-9764-837CAF96489E"),
+        CBUUID(string: "6E400003-B5A3-F393-E0A9-E50E24DCCA9E"),
     ]
 
     private static let pelagicGen1TxCharacteristic = CBUUID(
         string: "6606AB42-89D5-4A00-A8CE-4EB5E1414EE0"
     )
+    private static let crestCr5lTxCharacteristic = CBUUID(
+        string: "6E400002-B5A3-F393-E0A9-E50E24DCCA9E"
+    )
```
ericgriffin: The UUID `6E400002-B5A3-F393-E0A9-E50E24DCCA9E` is the standard Nordic UART Service RX characteristic - it's not CREST-specific, and plenty of non-CREST BLE peripherals use it.

Two consequences:
- the name `crestCr5lTxCharacteristic` is misleading (it's NUS, not CR5L; and it's RX from the peripheral's perspective, not Tx)
- the checks in `finalizeCharacteristicSelection` and `initialWriteWithoutResponsePreference` that key on this UUID will fire for any NUS peripheral, flipping it to write-with-response and multi-char subscription

If the intent is "make NUS work everywhere," great — suggest renaming to `nordicUartServiceRxCharacteristic` and updating the PR description. If the intent is truly CR5L-only, suggest gating on `peripheral.name?.hasPrefix("CREST")` so we don't change behavior for other NUS devices.

Also, a one-line comment explaining why CR5L needs write-with-response would help preserve the reverse-engineering context you gathered.

packages/libdivecomputer_plugin/darwin/Sources/LibDCDarwin/BleIoStream.swift
```diff
@@ -45,12 +51,14 @@ class BleIoStream: NSObject, CBPeripheralDelegate {
 
     private var writeCharacteristic: CBCharacteristic?
     private var notifyCharacteristic: CBCharacteristic?
+    private var subscribedCharacteristics: [CBCharacteristic] = []
```
ericgriffin, line 58: Minor: now that `subscribedCharacteristics` is the source of truth for which channels we're listening on, `notifyCharacteristic` (line 53) is still assigned and cleared but no longer used for gating. It only appears in the "Discovery ready" log. Either drop it or add a comment marking it as log-only, otherwise a future change that reads it will silently operate on only one of the N subscribed characteristics.

packages/libdivecomputer_plugin/darwin/Sources/LibDCDarwin/BleIoStream.swift
```diff
@@ -286,13 +295,15 @@ class BleIoStream: NSObject, CBPeripheralDelegate {
...
+            if partialReadBuffer.isEmpty, !readPackets.isEmpty {
+                partialReadBuffer = readPackets.removeFirst()
+            }
+            if !partialReadBuffer.isEmpty {
+                let bytesToRead = min(size, partialReadBuffer.count)
+                _ = partialReadBuffer.withUnsafeBytes { ptr in
                     memcpy(data, ptr.baseAddress!, bytesToRead)
                 }
-                readBuffer.removeFirst(bytesToRead)
+                partialReadBuffer.removeFirst(bytesToRead)
```
ericgriffin, line 325: Behavioral change worth documenting in the PR description: a single `performRead(size:)` now returns at most one packet's worth of bytes, even when multiple packets are already queued. Previously the flat `readBuffer` could fulfill `size` across packet boundaries in one call.

libdivecomputer's iostream contract allows short reads (via `actual`), so this should be fine, but it's a change for every device, not just CR5L. The Pelagic Gen1 path now does more read-loop iterations for large responses.

Not asking for a revert, just please mention in the PR description so future maintainers tracking down a throughput regression on another device can find this change easily.

packages/libdivecomputer_plugin/darwin/Sources/LibDCDarwin/BleIoStream.swift
```diff
@@ -626,7 +641,7 @@ class BleIoStream: NSObject, CBPeripheralDelegate {
                     + " \(error.localizedDescription)")
             return
         }
-        if let notify = notifyCharacteristic, characteristic.uuid != notify.uuid {
+        guard subscribedCharacteristics.contains(where: { $0.uuid == characteristic.uuid }) else {
```
ericgriffin, line 665: Before this PR, the guard was `if let notify = notifyCharacteristic, ... { return }` which meant a nil notify fell through. After, if `subscribedCharacteristics` is empty (pre-finalize, or a reset-during-reconnect race with in-flight notifications from a prior session), `contains(where:)` is false and every notification is silently dropped with no log.

Worth adding at least a warning so the drop is visible when debugging:

```swift
guard subscribedCharacteristics.contains(where: { $0.uuid == characteristic.uuid }) else {
    NativeLogger.w("BleIoStream", category: "BLE",
        "Ignoring notify from \(characteristic.uuid.uuidString) "
            + "(subscribed count=\(subscribedCharacteristics.count))")
    return
}
```

packages/libdivecomputer_plugin/darwin/Sources/LibDCDarwin/BleIoStream.swift
```diff
@@ -637,7 +652,7 @@ class BleIoStream: NSObject, CBPeripheralDelegate {
                 + " bytes=\(value.count)"
                 + " data=\(Self.hexString(value, maxBytes: Self.maxLogBytes))")
         readLock.lock()
-        readBuffer.append(value)
+        readPackets.append(value)
```
ericgriffin, line 683: `characteristic.value` can be a non-nil but empty `Data`. If an empty notification lands here, we append it to `readPackets` and signal the semaphore. On the next `performRead`, `partialReadBuffer = readPackets.removeFirst()` pulls the empty `Data`, the `!partialReadBuffer.isEmpty` branch doesn't fire, and we fall through to `readSemaphore.wait(...)` — but the signal was already consumed, so we sleep until the deadline even if more real data has since arrived.

Suggest guarding at the top of this callback:

```swift
guard let value = characteristic.value, !value.isEmpty else { return }
```

ericgriffin (commented):
Thanks for the thorough writeup and the gist links.

Overall the approach makes sense. A few things I'd like to flag before merge though:

1. Scope of the read-buffer change. The switch from `readBuffer: Data` to `readPackets: [Data]` + `partialReadBuffer` affects every device, not just CR5L — a `performRead` call can no longer span packet boundaries. I think that's fine under libdivecomputer's partial-read contract, but it's worth calling out explicitly in the PR description since it touches Pelagic Gen1 and anything else on the existing path.

2. "CR5L" behavior is actually keyed on the generic NUS RX UUID (`6E400002-B5A3-F393-E0A9-E50E24DCCA9E`), so any NUS-speaking peripheral would hit the multi-subscribe + write-with-response path. Might want to either narrow the gate (`peripheral.name` prefix) or rename/broaden the framing so it's honest about what's happening.

3. A couple of correctness issues in the state machine and notify handling (detailed inline) - one is a genuine silent-drop path I'd want to fix before merging.

Specifics inline. Happy to pair on any of these.

packages/libdivecomputer_plugin/darwin/Sources/LibDCDarwin/BleIoStream.swift
```diff
@@ -681,8 +695,11 @@ class BleIoStream: NSObject, CBPeripheralDelegate {
         }
         NativeLogger.d("BleIoStream", category: "BLE",
             "Notify enabled for \(characteristic.uuid.uuidString)")
-        isReady = true
-        signalDiscoveryReady()
+        if subscribedCharacteristics.allSatisfy(\.isNotifying) {
+            waitingForNotifyEnable = false
+            isReady = true
+            signalDiscoveryReady()
+        }
```
ericgriffin, line 736: The `waitingForNotifyEnable = false` used to fire unconditionally at the top of this callback; after this PR it's only cleared inside this `allSatisfy` branch.

In the multi-characteristic case, if one characteristic errors or comes back `!isNotifying`, we call `signalDiscoveryReady()` but leave `waitingForNotifyEnable` stuck at `true`. It mostly works out because `connectAndDiscover` resets state on entry, but any late/stray notification-state callback for another characteristic will still pass the `guard waitingForNotifyEnable else { return }` check and re-run the success logic with stale state.

Suggested fix: set `waitingForNotifyEnable = false` on both error branches above before signaling.

```swift
if let error {
    NativeLogger.e(...)
    waitingForNotifyEnable = false
    signalDiscoveryReady()
    return
}
if !characteristic.isNotifying {
    NativeLogger.w(...)
    waitingForNotifyEnable = false
    signalDiscoveryReady()
    return
}
```

While you're in there it might be helpful to include how many of `subscribedCharacteristics` are still pending in the error log, since otherwise the outer "Discovery completed without usable write/notify characteristics" message hides which of the N subscriptions failed.

packages/libdivecomputer_plugin/darwin/Sources/LibDCDarwin/BleIoStream.swift
```diff
@@ -749,23 +766,39 @@ class BleIoStream: NSObject, CBPeripheralDelegate {
...
+        if candidate.write.uuid == Self.crestCr5lTxCharacteristic {
+            subscribedCharacteristics = serviceCharacteristics.filter {
+                $0.properties.contains(.notify) || $0.properties.contains(.indicate)
+            }
+            if subscribedCharacteristics.isEmpty {
+                subscribedCharacteristics = [candidate.notify]
+            }
+        } else {
+            subscribedCharacteristics = [candidate.notify]
+        }
```
ericgriffin, line 840: Two concerns here:

**1.** `candidate.write.service?.characteristics` can be nil/empty in edge cases (cached peripheral, out-of-order discovery). The `?? []` + filter silently falls through to `[candidate.notify]`. Worth a warning log so the fallback is diagnosable:

```swift
if subscribedCharacteristics.isEmpty {
    NativeLogger.w("BleIoStream", category: "BLE",
        "NUS write char found but no notify/indicate chars on its service "
            + "(service=\(candidate.write.service?.uuid.uuidString ?? "nil")); "
            + "falling back to candidate.notify=\(candidate.notify.uuid.uuidString)")
    subscribedCharacteristics = [candidate.notify]
}
```

**2.** The scorer picks best write and best notify independently and may pair them across services. When we hit the fallback, `candidate.notify` could live on a service that isn't the write's service, which would be a silently-wrong subscription. Might be worth asserting `candidate.notify.service === candidate.write.service` for the CR5L/NUS path, or logging when they diverge.

ericgriffin (changes requested):

packages/libdivecomputer_plugin/darwin/Sources/LibDCDarwin/BleIoStream.swift
jefdriesen: The I/O semantics is different for packet based transports (BLE, USB and USBHID) and stream based transports (SERIAL, BLUETOOTH and IRDA):

For the packet based transports, the read function should always return **exactly one** packet. If the packet happens to be shorter than `size` bytes, the `actual` parameter should indicate the actual length of the packet. If the packet is larger than `size` bytes, your implementation should return an `DC_STATUS_IO` error to indicate the problem. You can optionally return the first `size` bytes of the packet, but that's mainly useful for debugging purposes.

Never return partial packets (without also returning an error), or concatenate data from multiple packets!

The reason is that some dive computers protocols use additional framing data (checksums, sequence numbers, etc) in their data packets. After validation, those framing bytes needs to be discarded to keep only the payload data. Libdivecomputer can only do that if the BLE packet boundaries are preserved.

For the stream based transports, there are no packets and the read function should simply try to read `size` bytes. If less bytes are available before the timeout expires, return a `DC_STATUS_TIMEOUT` error.

KalenJosifovski:
Thanks. I addressed the requested changes in 02a1eafc.

  Summary:
  - narrowed the NUS-specific behavior to likely CREST devices instead of keying only on the generic NUS RX UUID
  - kept `notifyCharacteristic` explicitly as log-only
  - added warning logs for ignored notifications and fallback paths
  - ignored empty notifications
  - fixed `waitingForNotifyEnable` on the error / !isNotifying paths
  - preserved the packet-based read behavior and updated the PR description accordingly

  On the read-path change: I also aligned it with libdivecomputer's packet-transport expectations from jefdriesen's comment. In particular, if a single queued BLE packet is larger
  than the requested read size, `performRead` now returns `LIBDC_STATUS_IO` instead of silently partial-reading it.

packages/libdivecomputer_plugin/darwin/Sources/LibDCDarwin/BleIoStream.swift
KalenJosifovski: Fixed in 02a1eafc. I renamed the UUID constants to generic NUS names and gated the NUS-specific multi-subscribe / write-with-response behavior behind a likely-CREST device check (`isLikelyCrestDevice`). I also added a comment in `initialWriteWithoutResponsePreference` noting that if someone wants to broaden support for more NUS devices later, that is the place to widen the gate.

packages/libdivecomputer_plugin/darwin/Sources/LibDCDarwin/BleIoStream.swift
KalenJosifovski: Fixed in 02a1eafc. I kept `notifyCharacteristic`, but marked it explicitly as log-only so it is not mistaken for the active subscription source of truth. `subscribedCharacteristics` remains the actual gating state.

packages/libdivecomputer_plugin/darwin/Sources/LibDCDarwin/BleIoStream.swift
KalenJosifovski: Updated in 02a1eafc and documented in the PR body. I am keeping the packet-preserving behavior here. Jef's follow-up comment matches the libdivecomputer BLE contract more precisely: packet transports should return exactly one packet, not concatenate multiple packets into a stream. I also tightened the oversized-packet edge case so if one queued BLE packet is larger than the requested read size, `performRead` now returns `LIBDC_STATUS_IO` instead of silently splitting it.

packages/libdivecomputer_plugin/darwin/Sources/LibDCDarwin/BleIoStream.swift
KalenJosifovski: Fixed in 02a1eafc. Ignored notifications now log a warning with the characteristic UUID and current subscribed count so this path is visible during debugging.

packages/libdivecomputer_plugin/darwin/Sources/LibDCDarwin/BleIoStream.swift
KalenJosifovski: Fixed in 02a1eafc. I added an empty-data guard (`guard let value = characteristic.value, !value.isEmpty else { ... }`) so empty notifications are ignored instead of enqueueing
  an empty packet and consuming the read semaphore incorrectly.

packages/libdivecomputer_plugin/darwin/Sources/LibDCDarwin/BleIoStream.swift
KalenJosifovski: Fixed in 02a1eafc. `waitingForNotifyEnable` is now cleared on both failure branches before signaling discovery readiness. I also included the remaining pending count in those
  logs so it is clearer which subscription phase failed.

packages/libdivecomputer_plugin/darwin/Sources/LibDCDarwin/BleIoStream.swift
KalenJosifovski: Fixed in 02a1eafc. I added warning logs for both edge cases: when the write characteristic service has no notify/indicate characteristics and we fall back to `candidate.notify`, and when the selected write and notify characteristics come from different services on the Crest/NUS path.

KalenJosifovski:
Two extra points for context:

  - I’ll leave the review threads for @ericgriffin to resolve as maintainer once he’s happy with the
  follow-up commit.
  - I don’t have the CR5L watch with me right now and won’t have it again until Monday, so I have not been
  able to re-run a live physical-device test after this review-response commit. I’m happy to hold off on
  asking for merge until I can retest against the watch if that’s preferred.

packages/libdivecomputer_plugin/darwin/Sources/LibDCDarwin/BleIoStream.swift
```diff
@@ -286,13 +295,15 @@ class BleIoStream: NSObject, CBPeripheralDelegate {
 
         while true {
             readLock.lock()
-            let available = readBuffer.count
-            if available > 0 {
-                let bytesToRead = min(size, readBuffer.count)
-                _ = readBuffer.withUnsafeBytes { ptr in
+            if partialReadBuffer.isEmpty, !readPackets.isEmpty {
+                partialReadBuffer = readPackets.removeFirst()
+            }
```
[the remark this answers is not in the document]
ericgriffin: Not a performance concern and doesn't need to be addressed.

packages/libdivecomputer_plugin/darwin/Sources/LibDCDarwin/BleIoStream.swift
```diff
@@ -16,19 +16,25 @@ class BleIoStream: NSObject, CBPeripheralDelegate {
 
     private static let preferredServiceUUIDs: Set<CBUUID> = [
         CBUUID(string: "CB3C4555-D670-4670-BC20-B61DBC851E9A"),
+        CBUUID(string: "6E400001-B5A3-F393-E0A9-E50E24DCCA9E"),
     ]
 
     private static let preferredWriteUUIDs: Set<CBUUID> = [
         CBUUID(string: "6606AB42-89D5-4A00-A8CE-4EB5E1414EE0"),
+        CBUUID(string: "6E400002-B5A3-F393-E0A9-E50E24DCCA9E"),
     ]
```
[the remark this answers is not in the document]
ericgriffin: Addressed in 02a1eafc

ericgriffin (approved):
👍 Let me know once you're able to test again and then happy to merge. Thanks.

KalenJosifovski:
Quick update from additional live testing.

I hit and fixed a new robustness issue locally that is not in this PR branch yet:
- CR5L scan alias matching (`CREST-CR5L` advertisement name variants)
- known-device vendor/product normalization before descriptor lookup

Those fixes are currently in my local `cr5l-libdc-log-wiring` branch. I can port them into this PR as a small focused follow-up commit if you want.

Separately, I also wired explicit `libdivecomputer` internal log callback handling in that same branch for protocol debugging. That logging wiring is useful for diagnostics, but I’m treating it as optional and can keep it out of the main PR unless you want it included.

Related context: I’m currently working with Jef on the libdivecomputer PR to settle final CR5L protocol framing details using the raw capture. Since this Submersion support depends on that backend behaviour, it may be cleaner to let that resolve first and then finalise here. No rush on my side.

ericgriffin:
Thanks for the update. Feel free to bring all of the things you mentioned into this branch if you'd like to. Happy to merge it whenever you feel it's ready.

KalenJosifovski:
Thanks Eric, will do. I will try and get this done before I go on annual leave end of May but cant promise anything. Either way we will be waiting on the libdivecomp side so no rush I guess.

ericgriffin:
Resolved the conflicts with current `main`, but this fork has **Allow edits by maintainers** turned off, so the result could not be pushed to `cr5l-support` (`remote rejected ... permission denied`). The merged branch is published on the main repo as [`cr5l-support-merged-207`](https://github.com/submersion-app/submersion/tree/cr5l-support-merged-207) instead.

To bring it into this PR, either enable maintainer edits on the PR, or pull it yourself:

```
git remote add upstream https://github.com/submersion-app/submersion.git
git fetch upstream cr5l-support-merged-207
git checkout cr5l-support
git reset --hard upstream/cr5l-support-merged-207
git push --force-with-lease origin cr5l-support
```

**What the merge had to decide.** `BleIoStream.swift` had 13 conflicting hunks, because `main` has rewritten most of what this PR also touches. Interleaving two rewrites of a BLE transport is not something to attempt hunk by hunk, so `main`'s file was taken whole and this PR's actual contribution re-expressed against its structure.

*Superseded by main, so dropped:*
- The `readPackets` / `partialReadBuffer` rework. `main` solved the same problem more thoroughly in `PacketReadBuffer`, which returns bytes from at most one BLE notification per read (issue #394). Your diagnosis was right; it just landed from the other direction.
- The characteristic scoring inline in `BleIoStream`. `main` extracted it into `BleCharacteristicSelector`, a CoreBluetooth-only value-type module with a standalone Swift test target.

*Ported forward, which is what this PR is for:*
- The Nordic UART UUIDs and the Crest-gated preference for them, now in `BleCharacteristicSelector`. `select()` takes the peripheral name and biases the NUS service, RX and TX only when the name says Crest. Your reasoning for keeping this name-gated rather than global is preserved verbatim in the code comment: NUS is a generic Nordic SDK service and biasing it unconditionally could pull an unrelated device off its vendor service.
- The acknowledged-write preference for NUS RX, which advertises both write modes and would otherwise be claimed by the `writeWithoutResponse` bias.

**One thing deliberately not ported, and I would like your read on it.** The multi-characteristic subscription (`subscribedCharacteristics`) is gone. `main` now pins the notify characteristic by preferred UUID, so NUS TX is selected directly and subscribing to every notify characteristic on the service is no longer needed to *find* it. Reworking `main`'s credits-then-data `notifySetupStep` into a set would put the shipping Telit and u-blox (OSTC) paths at risk. If your CR5L actually needed the broader subscription for a reason beyond discovery, say so and it goes back in.

**Verified:** all Darwin native Swift tests pass, including five new `BleCharacteristicSelector` cases. Two of those are the ones that matter for everyone else: a non-Crest peripheral and an unnamed peripheral select exactly as they did before the gate existed. macOS Debug build succeeds. Not tested against real CR5L hardware, which I do not have.

ericgriffin:
Will re-open in the future if https://github.com/libdivecomputer/libdivecomputer/pull/59 is ever merged.

PR #342, opened 2026-06-16
feat(import layer): improve parsing of Cressi Leonardo download

essengee:
## Summary

  Improves dive data parsing for the original Cressi Leonardo dive computer. Previously, all ascent rate indicators from
  the Leonardo were misclassified as "Deco Violation" alerts due to two bugs: an off-by-one error in the event type
  mapper (all platforms), and the parser emitting events for non-alarm ascent levels. Additionally, the Leonardo's deco
  obligation flag (stored as bit 11 in the sample data) was never extracted, meaning neither deco status nor deco
  start/end events were reported. Water temperature from the dive header was also not being saved to the dive record.

## Changes

  libdivecomputer parser (cressi_leonardo_parser.c)

  - Extract bit 11 from 16-bit sample word as deco obligation flag
  - Emit DC_SAMPLE_DECO with DC_DECO_DECOSTOP or DC_DECO_NDL on every sample based on bit 11
  - Emit SAMPLE_EVENT_DECOSTOP with BEGIN/END flags on deco state transitions (0→1 and 1→0)
  - Only emit SAMPLE_EVENT_ASCENT when ascent rate level = 3 (the only level that triggers an audible alarm on the
  Leonardo hardware)

  Event type mapper fix (all platforms)

  - Fixed off-by-one in MapEventType / mapEventType — was missing SAMPLE_EVENT_RBT (value 2), causing all event types ≥2
  to map to the wrong string
  - Affected: Windows (dive_converter.cc), Linux (dive_converter.c), Android (DiveComputerHostApiImpl.kt), iOS/macOS
  (DiveComputerHostApiImpl.swift)

  Dart import layer

  - _mapEventTypeString now accepts optional flags parameter to distinguish deco start (flags=1) from deco end (flags=2)
  - importProfile accepts minTemperature/maxTemperature parameters and writes waterTemp to the dive record (fixes
  Leonardo temperature not appearing)
  - Same changes applied to reparse service for consistency

  Tests

  - Added 5 unit tests covering: temperature import from header, ascent event classification, deco start/end flag
  handling, and ceiling violation mapping
  - Regenerated mocks for new importProfile signature
## Test Plan

- [x] `flutter test` passes
- [x] Manual testing on: Windows

## Screenshots

**Note:** NDL and ceiling in the Submersion screenshots below are calculated at 50/75 Buhlmann gradient - as close as I could get to the RGBM calculation of the computer for this dive. The computer itself does not include this level of detail in its 16 bits of sample data at 20 second intervals. Working with old tech here...

Result of original import:
CressiLeonardoOriginalImport

Result of import after changes:
CressiLeonardoImprovedImport

Screenshot of (ancient) Cressi PC Interface display of same dive:
CressiPCInterfaceDisplay


essengee:
Hi @ericgriffin !

I've been wanting to work on this for some time and finally decided to day to give it a go. It (unfortunately) involves a change to libdivecomputer code and I'm not sure how (or even if!) you want to handle that; I don't have a fork of the submersion_patches but I can certainly figure that out if necessary. Just wasn't sure if it was better to do it ham-fisted or let you handle it since I think you by far know what you are doing more than I do in this area.

ericgriffin:
Merged current `main` into this branch to clear the conflicts. Three things here were not ordinary merge noise, so recording what was done:

**1. The libdivecomputer submodule pointer was unreachable.** The branch pointed at `5ee8354a` in `essengee/libdivecomputer`, which now 404s, so `git submodule update` failed for anyone but you. The branch's own `patches/0002-cressi-leonardo-deco-and-ascent-fix.patch` was the only surviving record; it applies cleanly to the current `submersion-patches` tip, so the commit was rebuilt on top of it as `9757d079` and is now published at `submersion-app/libdivecomputer` on branch `cressi-leonardo-342`. The submodule pointer resolves for everyone again. (`submersion-patches` itself is untouched; it can be fast-forwarded when this merges.)

**2. The patch number collided.** Main's series grew to 0001-0007 and `0002` is now `hw-ostc3-ble-partial-read`, so the Cressi patch is renumbered to `0008-cressi-leonardo-deco-and-ascent-fix.patch`.

**3. The event-table renumbering is kept, deliberately.** This PR inserts `rbt` at index 2 in the sample-event name table and shifts everything after it, across all four bridges (Kotlin, Swift, `linux/dive_converter.c`, `windows/dive_converter.cc`). Checked against `include/libdivecomputer/parser.h`: `SAMPLE_EVENT_RBT` really is index 2, so **main is currently off by one from index 2 upward**, mislabelling ascent as deco, ceiling as ascent, and so on. This is a real bug fix in main, not a regression, and main's table was not restored.

The four `*.mocks.dart` conflicts resolve to main's deletion: Mockito output moved to build_runner and `*.mocks.dart` is gitignored now, so those files are regenerated from the same `@GenerateMocks` annotations and nothing is lost.

Verified locally: `flutter analyze` clean, 2,689 tests green across `test/features/dive_computer`, `test/features/import_wizard` and `test/features/universal_import`.

ericgriffin:
Thanks for this, and sorry it sat so long. Both of your findings about the
Leonardo held up under review and are shipping in #1527, which relands this
work with your authorship on the libdivecomputer commit.

Closing here rather than merging, because the branch is now 900+ commits
behind and a large part of it has been overtaken:

**Kept, and both correct.** Bit 11 of the sample word really is the deco
obligation, and gating `SAMPLE_EVENT_ASCENT` to level 3 is the right call. An
ascent event means the computer raised an alarm, so emitting one for the
levels that only move the rate indicator was wrong regardless of how the app
renders it.

**Superseded.** The event-code table off-by-one you found was a genuine bug in
all five bindings. It was fixed independently on 2 September with one shared
`libdc_event_type_name()` in `macos/Classes/libdc_download.c`, a Kotlin twin
pinned by a test, and a native test that walks the whole enum. Merging this
branch's version would have reverted that consolidation, so #1527 leaves the
four binding files alone. That fix also already resolves the headline symptom
here: ascent events no longer show as Deco Violation.

**Changed.** The patch reported `DC_SAMPLE_DECO` on every sample. Since June
the wrapper gained deco carry-forward across samples (`libdc_download.c`,
`current_deco_type` seeded to `UINT32_MAX` for "never reported"), so #1527
reports only on transitions. That matters for more than tidiness: with
per-sample emission, `DC_DECO_NDL` with `time = 0` reaches
`parsed_dive_mapper.dart` as `ndl: s.decoType == 0 ? s.decoTime : null`, so
every Leonardo dive would come back claiming zero no-stop time for its whole
length, and `profile_analysis_provider.dart` would then offer that flat zero
as the computer's own NDL curve in place of the calculated one. Same for the
stop depth and the ceiling. #1527 reports nothing on a no-deco dive, leaving
the obligation unknown, and adds the zero-series rule that TTS already had.

Two smaller adjustments: the patch is gated to the Leonardo model id, because
the Giotto, Newton and Drake share this parser and neither bit 11 nor the
ascent grading is confirmed on them; and the water-temperature fix coalesces
with the sample-derived minimum rather than overwriting it, which would
otherwise have blanked the water temp of computers that report temperature
only in the sample stream.

Also worth recording: the submodule pointer on this branch would have reverted
six libdivecomputer commits, including the Seac Tablet BLE name filter (#1419)
and the Ratio iX3M 2 APOS5 record-type fix (#1481). Nothing in the diff showed
it, because a gitlink change has no file content. #1527 adds a CI job that
blocks any PR moving the pointer to a commit that does not contain the base
branch's, which is the general fix for a class of regression review cannot
catch.

ericgriffin:
Superseded by #1527, which relands this work. Closing.

PR #350, opened 2026-06-18
fix(secure-storage): legacy-keychain fallback never switched keychains on macOS (S3 -34018)

ericgriffin:
## Summary

Follow-up to #348 (merged). With the no-sandbox DMG launching again, S3 sync on it failed with `errSecMissingEntitlement` (-34018) — and after the first cut of this fix, credentials *saved* but came back **blank** on reload. Same underlying issue: the no-sandbox build can't use the macOS data-protection keychain, and #340's legacy-keychain fallback never actually worked.

## Two bugs in #340's fallback

1. **The keychain switch was a silent no-op (Dart/native key mismatch).** `MacOsOptions(usesDataProtectionKeychain: false)` is serialised by Dart under key `usesDataProtectionKeychain`, but `flutter_secure_storage_darwin` 0.2.0 reads `useDataProtectionKeyChain` and defaults to `true` — so the legacy keychain was never selected. (Upstream bug; worked around with a `MacOsOptions` subclass that emits the native-read key.)

2. **The fallback was asymmetric.** It triggered only on -34018, which only *modifying* ops raise. A keychain **read** of a missing item returns not-found (`null`), not -34018, so reads never fell back: writes went to the legacy keychain, but reads queried the empty data-protection keychain → `null`. Result: creds saved, then "not authenticated" / blank fields.

## Fix — probe once, then route

Replace the per-op error fallback with a one-time probe: a throwaway write to the data-protection keychain raises -34018 iff it's unavailable. Cache the verdict and route **every** read/write/delete to the keychain that works:
- no-sandbox build → legacy file-based keychain (reads now find what writes stored, including the cross-launch first read at sync init),
- sandboxed App Store / iOS → data-protection keychain (and it never touches the file-based keychain the sandbox forbids).

## Test plan

- [x] `flutter test` secure-storage + S3/bookmark suites — 22/22 green. The fakes now model the keychain's real asymmetry (data-protection reads return `null`, not -34018), so the new **write-then-read round-trip** test fails on the old design and passes on this one.
- [x] End-to-end on the no-sandbox build: configure S3, Save, reopen — creds persist, sync authenticates.

## Upstream

The Dart/native key-casing mismatch is a `flutter_secure_storage` bug; reporting upstream. The `MacOsOptions` subclass workaround can be removed once the keys agree.

### DIFF
--- a/lib/core/services/secure_storage/fallback_secure_storage.dart
+++ b/lib/core/services/secure_storage/fallback_secure_storage.dart
@@ -1,19 +1,26 @@
+import 'package:flutter/foundation.dart' show visibleForTesting;
 import 'package:flutter/services.dart';
 import 'package:flutter_secure_storage/flutter_secure_storage.dart';
 
-/// A thin wrapper over [FlutterSecureStorage] that transparently retries on
-/// the macOS legacy (file-based) keychain when the default data-protection
-/// keychain reports `errSecMissingEntitlement` (-34018).
+/// Wrapper over [FlutterSecureStorage] that uses the macOS legacy (file-based)
+/// keychain when the default data-protection keychain is unavailable.
 ///
-/// That status occurs on the ad-hoc-signed no-sandbox (GitHub-distribution)
-/// build, whose signature carries no team `application-identifier` and hence
-/// no keychain access group. Sandboxed / team-signed / iOS builds never hit
-/// it, so the first attempt -- byte-identical to a plain [FlutterSecureStorage]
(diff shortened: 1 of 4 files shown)


ericgriffin:
Verified end-to-end on a local no-sandbox build: S3 credentials **save**, **persist** when reopening the S3 screen, and **sync authenticates** (no `-34018`). The probe correctly pins to the legacy keychain. The remaining test-plan item is confirmed. ✅

Issue #276, opened 2026-05-04
SAC rate disappears after edit dive (setting dive location, trip, operator etc)


jonsedar:
### Background:
+ MacOS Sequoia: 15.7.5
+ Submersion: Version 1.4.6 (92)
+ Imported dives from a subsurface export: 6.0.5576, file type SSRF (XML), originally captured on a Perdix 2 with Swift AI. SAC rate is available.
+ View list of newly imported dives in Submersion, see SAC rate graph in the profile panel (see attached screenshot "pre")

<img width="1522" height="407" alt="Image" src="https://github.com/user-attachments/assets/eedbf0ec-234e-4a84-b228-14c4b56e98ff" />

### Action:
+ Edit dive and update something:
+ To date, I've seen this happen when changing
    + dive site name
    + trip name
    + operator name

### Expected:
+ Dont lose any info

### Actual:
+ Lose SAC rate info in the profile, also no longer selectable in the details dropdown (see attached screenshot "post")

<img width="1516" height="414" alt="Image" src="https://github.com/user-attachments/assets/5ea3bfe3-f4b7-4160-ba11-e380b147d7d1" />

### Related issues (possible):
+ https://github.com/submersion-app/submersion/issues/191

### Also Note:
+ If I import the Perdix straight into submersion, the SAC info is also lost
+ I suspect this could be an issue where SAC is not getting recalculated following an edit, or the reference to a precalculated SAC is getting deleted or un-keyed

PR #405, opened 2026-06-24
Apply Z-factors to SAC/RMV calculations

### LINKED ISSUE #379: SAC/RMV calculations result in values higher than Subsurface
Comparing the calculated SAC rate in Submersion with the rate reported in Subsurface for dives imported from my non-air integrated computer, the Submersion rate is consistently ~1 L/min higher, sometimes a bit more:

Subsurface:
Image

Submersion:
Image

Comparing Submersion source to the source from a recent fork of Subsurface it appears the difference is due to Submersion using ideal gas laws in calculations, whereas Subsurface applies Z factors to take into account compressibility.

Is this by design? If not, I have some work in a local branch that applies the Z factors. With the changes, all of the SAC calculations in Submersion are within 0.2 L/min of the Subsurface ones, for dives at varying depths:
Image

essengee:
## Summary

When using SAC/RMV calculation in the app instead of values provided by dive computers, the calculation provides results significantly higher than what is observed for the same dive profiles and start/end tank pressure inputs in Subsurface. Comparing the open-source Subsurface code to Submersion indicated that Submersion was using ideal gas laws in its calculations instead of applying the industry-standard Z-factors to take gas compressibility into account. This PR applies those factors to more closely match expected real-world SAC/RMV rates.

## Changes

- applies Z-factors when calculation SAC/RMV rates per-dive
- applies Z-factors when calculating five-year SAC/RMV statistics
- tests included for the above: 100% coverage of changes based on flutter test --coverage

## Test Plan

- [x] `flutter test` passes
- [x] Manual testing on Windows: compared SAC/RMV calculation results for ~40 dives in Submersion with the values shown in Subsurface for the same imported dives. Submersion is now within +/- 0.2 L/min of Subsurface calculations; likely the result of slightly different rounding


ericgriffin:
### Code review

Found 2 issues:

1. Per-cylinder SAC silently becomes `null` for tanks that have pressures but no recorded volume. In `calculateCylinderSac`, the non-time-series branch now gates the computation on `tank.volume != null` with no pressure-based fallback, so a tank with start/end pressures but no volume and no tank-pressure time series returns `sacRate = null` — previously it returned a bar/min value. This silently reverts commit `814c553c7b5` ("show SAC on graph for imports without tank volume") and its #276 follow-up `aa2557b6772` ("SAC is bar/min, so volume is not required"); tank volume is commonly absent on imports. (regression / historical git context)

https://github.com/submersion-app/submersion/blob/02145fd290566ab443cff24321959825e2a8dfb7/lib/features/dive_log/data/services/gas_analysis_service.dart#L290-L296

2. `_zCorrectedSacRate` returns atm/min, but its doc comment and the `CylinderSac.sacRate` contract are bar/min. `gasUsedLiters / tankVolume` reduces to `(P / Z) / standardAtmBar` because [`gasVolume()`](https://github.com/submersion-app/submersion/blob/02145fd290566ab443cff24321959825e2a8dfb7/lib/core/utils/gas_compressibility.dart#L41-L55) already divides by `standardAtmBar` (1.01325) and returns "liters at 1 atm". The no-volume fallback returns raw `pressureUsed` in true bar/min, so the same `sacRate` field differs by ~1.3% depending only on whether `tank.volume` is set. (unit bug — doc says "Returns the SAC rate in bar/min at surface")

https://github.com/submersion-app/submersion/blob/02145fd290566ab443cff24321959825e2a8dfb7/lib/features/dive_log/data/services/gas_analysis_service.dart#L336-L362

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

ericgriffin (approved):

PR #435, opened 2026-06-28
Spec for gas-aware ascent in deco calculations

readme42:
Design for Variant 1 from Discussion #326: make the simulated ascent (TTS/ceiling/stop schedule) select the best available gas at each simulated stop depth via a pluggable AscentGasPlan strategy, instead of projecting the whole ascent on the single currently-active gas. Keeps the bottom-phase tissue-loading path unchanged and adds a dedicated CCR setpoint+diluent ascent model.


ericgriffin (approved):
👍

PR #448, opened 2026-07-02
perf(sync): restore orphaned S2 batched-writes + S3 isolate parse (large-library sync speed)

ericgriffin:
## Summary

Restores the sync **performance** work (S2 batched writes + S3 isolate parse) that was **merged but orphaned from `main`**, and re-applies it cleanly on top of #447.

**Stacked on #447** (base = `worktree-issue-358-publish-base-oom-streaming`) — this PR's diff is only the S2/S3 commits. Merge #447 first, then rebase this onto `main`.

### Why this is needed
PRs **#417 (S2)** and **#415 (S3)** merged 2026-06-25 but are **not in the current `main`** (`708425c0cc8`): the history diverged at #412 into a perf lineage and a feature lineage, and `main` is the feature lineage. Current `main` has **0 `upsertRecords`**, no isolate, and still row-by-row `_mergeEntity` — which is why syncing a large library is slow (~10-20 min) and the progress spinner freezes.

### What this restores
- **S2 — batched pull writes:** `_mergeEntity` read-decide-write (`fetchRecords` + `upsertRecords`) instead of a per-row upsert loop. Fixes the **slow sync**. (`upsertRecords` itself already lives in #447, which ported it for the adopt path; this reuses it — no duplication.)
- **S3 — isolate base-parse:** `BaseParseClient`/`base_parse_worker` offload the base-file parse+apply to a worker isolate, with an **inline fallback** if the isolate can't spawn. Fixes the **UI-thread freeze** (spinner stalls).

Together with #447's batched *adopt* path, all three large-library sync paths (pull, adopt, publish) are now batched and off the main isolate.

### Verification
- Full suite: **9,492 passed / 12 skipped / 0 failed**; whole-project analyze + format clean.
- Cherry-picked S2/S3 commits (history preserved) with their tests, incl. `base_parse_client_test` and the batched-serializer tests; the isolate's inline-fallback path is tested.
- Adopt + base-streaming parity tests stay byte-for-byte green.

### Notes
- S1 (stream-coalescing / UI-rebuild debounce) from the orphaned lineage is **not** included here — separate, non-blocking.
- Redundant coverage: `sync_data_serializer_batch_test.dart` (from S2) and `sync_data_serializer_batch_coverage_test.dart` (from #447) overlap on `upsertRecords`; harmless, left as-is.

### DIFF
--- a/lib/core/services/sync/changeset_log/base_parse_client.dart
+++ b/lib/core/services/sync/changeset_log/base_parse_client.dart
@@ -1,7 +1,7 @@
 import 'dart:async';
 import 'dart:isolate';
 
-import 'base_parse_worker.dart';
+import 'package:submersion/core/services/sync/changeset_log/base_parse_worker.dart';
 
 /// Thrown when the base-parse worker reports a parse/checksum error or dies.
 class BaseParseException implements Exception {
@@ -13,8 +13,8 @@ class BaseParseException implements Exception {
 
 /// Main-isolate client for the base-file parse worker. Spawns a worker that
 /// reads + parses a serialized sync base document off the UI isolate and streams
-/// decoded rows back, pull-backpressured. See [baseParseWorkerMain] for the wire
-/// protocol. Operations are sequential (one in flight at a time).
+/// decoded rows back, pull-backpressured (one ≤500-row batch per [nextDataBatch]).
+/// Operations are sequential — one in flight at a time.
 class BaseParseClient {
   BaseParseClient._(this._isolate, this._toWorker, this._fromWorker, this._sub);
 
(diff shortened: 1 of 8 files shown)


ericgriffin:
Addressed both review comments in `d34b8cb`:

- **fetchRecords `IN (...)` overflow:** `fetchRecords` now chunks the id list (900/slice, recursing to the per-entity switch) and merges results, so a large changeset apply can't exceed SQLite's ~999 bound-variable limit. Test: 1500-id fetch.
- **Isolate hang:** `BaseParseClient._nextMessage` now has a bounded per-message timeout that throws `BaseParseException`; `_applyRemoteBaseFile` already catches it and falls back to the inline parser, so a worker that dies/hangs after the handshake (no `onError`) can't hang the sync. Test: ready-then-silent worker.

Full sync suite green (362); whole-project analyze + format clean.

PR #454, opened 2026-07-02
fix(deco): keep TTS mandatory-only, report the safety stop separately

readme42:
## What this changes

Makes the **time-to-surface (TTS)** shown on the dive profile count only the **mandatory** obligation -- the ascent plus any required decompression stops -- and reports the optional safety stop **separately** instead of folding it into TTS.

## Why it was wrong before

TTS baked a recommended 3-minute safety stop into the no-deco ascent time. As soon as a dive crossed from "no deco" into a real decompression obligation, that safety stop was absorbed into the (now mandatory) deco stops, and the displayed TTS actually **dropped** -- e.g. 8 min to 5 min -- right at the NDL to deco transition. That reads backwards: the diver just picked up an obligation, so TTS should never fall there.

## What it does now

- **In deco** (`ndl < 0`): TTS = ascent + required deco stops. The mandatory stops supersede any recommended safety stop, so no safety stop is reported.
- **No deco** (`ndl >= 0`): TTS = direct ascent to the surface only. The safety stop is reported separately on `DecoStatus.safetyStopSeconds` (the 3-minute recommendation minus time already spent in the safety-stop zone).

TTS can now only ever rise as an obligation appears -- never drop as the diver enters deco. The safety stop is not lost: the profile already surfaces performed safety stops as their own phase/event, and the value is now available on
`DecoStatus` for display.

## Screenshot before:

Note the drop of the TTS curve from 8 minutes to 5 minutes at the moment when deco obligation starts.
Also note that TTS is 3 minutes during the very beginning of the descent as the safety stop was included in TTS.

image

## Screenshot after:

TTS rises during descent, stays the same until NDL is 0 and then increases with deco obligations.

image

### DIFF
--- a/lib/core/deco/buhlmann_algorithm.dart
+++ b/lib/core/deco/buhlmann_algorithm.dart
@@ -463,35 +463,31 @@ class BuhlmannAlgorithm {
         ? calculateDecoSchedule(currentDepth: currentDepth, fN2: fN2, fHe: fHe)
         : <DecoStop>[];
 
-    // Calculate TTS - even for no-deco dives, include safety stop time
-    // so divers know realistic ascent time.
+    // TTS is the MANDATORY time to surface only: the ascent plus any required
+    // decompression stops. The recommended safety stop is NOT folded in here --
+    // it is reported separately in safetyStopSeconds. Baking the safety stop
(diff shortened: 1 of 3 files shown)


ericgriffin (approved):
👍

PR #473, opened 2026-07-04
feat(sync): Google Drive sync backend (all platforms, PKCE desktop)

ericgriffin:
## Summary

Makes Google Drive a complete, fully functional cloud-sync backend on all five platforms (iOS, macOS, Android, Windows, Linux). The Drive v3 appDataFolder REST layer pre-existed but was hidden and desktop-incapable; this PR adds the auth machinery, re-exposes the settings tile, and configures OAuth per platform.

Built via an authenticator-seam design (Approach A): one `GoogleDriveStorageProvider` keeps all Drive REST logic, with a `GoogleDriveAuthenticator` seam (boundary is `http.Client`) backed by two implementations — `GoogleSignInAuthenticator` (iOS/macOS/Android, `google_sign_in` v7) and `DesktopOAuthAuthenticator` (Windows/Linux, loopback OAuth via `googleapis_auth`).

Design spec and implementation plan: `docs/superpowers/specs/2026-07-02-google-drive-sync-backend-design.md`, `docs/superpowers/plans/2026-07-02-google-drive-sync-backend.md`.

## Changes

- **Auth seam** — `GoogleDriveAuthenticator` interface; `google_sign_in` logic extracted into `GoogleSignInAuthenticator` (with Android `serverClientId`); new `DesktopOAuthAuthenticator` doing PKCE loopback OAuth with refresh-token persistence in `GoogleDriveTokenStore` (keychain via `FallbackSecureStorage`, mirroring the S3 store).
- **Provider refactor** — `GoogleDriveStorageProvider` delegates auth to the seam; adds a one-shot 401 silent-re-auth retry and quota/offline error mapping; `_findFile` now rethrows auth errors instead of masking them as "not found" (which could fork a duplicate on token expiry).
- **Desktop uses PKCE, no client secret** — Google lists `client_secret` as optional for Desktop-app clients, so the loopback token exchange authenticates with the PKCE `code_verifier` alone. Only public OAuth **client IDs** are committed; nothing for secret scanning to flag.
- **UI** — restored the Google Drive tile in Cloud Sync settings (removed the `googledrive → null` coercion), with account-email subtitle, availability gating, and a cancellable browser-wait dialog on desktop. Three new l10n strings translated across all 11 locales.
- **Availability** — `googleDriveAvailableProvider` / `googleDriveAccountEmailProvider`; `GoogleDriveClientConfig.isSupportedOnThisPlatform` is the single source of truth shared with the `supportsGoogleDrive` capability flag so they can't diverge.
- **Platform OAuth config** — macOS `GIDClientID` + reversed URL scheme + Keychain Sharing entitlement; Android + Desktop OAuth clients registered (project `433819313354`, shared appDataFolder).

## Test Plan

- [x] `flutter test` passes (9,624 passed, 14 platform-gated skips)
- [x] `flutter analyze` passes (whole project, clean)
- [ ] Manual testing on: iOS, macOS, Android, Windows, Linux — per `docs/superpowers/specs/2026-07-02-google-drive-sync-manual-test-checklist.md` (real-device OAuth can't run on CI). This is the remaining pre-merge gate.

Two desktop-dialog widget tests are `skip`ped on macOS and execute for the first time on Linux CI (inverse of the existing Apple-only tile-tap pattern) — worth watching the first CI run.

## Merge note

A recent Dropbox sync provider (PR #461) touches the same files (`cloud_sync_page.dart`, `sync_providers.dart`, `storage_providers.dart`, the `CloudProviderType` enum, ARB files). If a `dropbox` enum value has landed on `main`, the exhaustive `switch` in `cloudProviderInstanceFor` won't compile until the merge adds the `dropbox` case — the analyzer catches it immediately; a mechanical merge fix, not a runtime hazard. ARB/tile changes merge additively.


ericgriffin:
Addressed the review feedback and resolved the merge conflict with `main`.

**Merge with `main` (Dropbox provider #461):** the only real conflict was a duplicate import in `cloud_sync_page.dart` (kept both); the `CloudProviderType` switch and provider tiles auto-merged with all four providers. Regenerated Drift/mockito codegen for `main`'s new tables (checklists, buddy roles, trips, consolidation). Full suite green: 10,142 passed / 14 platform-gated skips, `flutter analyze` clean.

**Copilot comments (all in `1d5ba305e5f`):**
- **Revocation token not URL-encoded** — fixed with `Uri.encodeQueryComponent`; added a test that a `1//…`-style token is percent-encoded in the body.
- **Cancel-vs-completion dialog race** (both dialog comments) — restructured `_authenticateWithBrowserWait` so auth-completion and Cancel both route through a single `closeDialog` guard that is checked-and-set synchronously (no read-after-`await`), so exactly one pop happens and it can only target the dialog route.
- **`hasProvider` ignored Drive availability** — Sync Now is now gated on `googleDriveAvailable` for a persisted Google Drive selection; added a widget test covering the unavailable-build case.

Desktop OAuth was also changed to PKCE with no committed client secret (Google lists `client_secret` as optional for Desktop-app clients), so nothing sensitive is in the repo. Real-device sign-in across the five platforms remains the pre-merge gate per the committed manual checklist.

jamespet77:
what is this change waiting on? I pulled the change and merged it with main and it appeared to be working after a few small updates because the working tree has changed so much.  I only ask because I would prefer using google drive.

ericgriffin:
Superseded by #1105.

This branch was cut on 2026-07-04 and had fallen 3,262 commits behind `main`, with conflicts in the provider, the Cloud Sync page, both provider files, and all 11 ARB files. Rather than rebase it across a provider that both sides had independently refactored, #1105 was cut fresh from `origin/main`: the self-contained pieces (the authenticator seam, desktop PKCE loopback, token store, client config, and their tests) carried over verbatim, and the integration points were reapplied by hand.

Three of those integration points would have merged cleanly but misbehaved at runtime, which is why the port was done by hand:

- `debugSetDriveApi`, added to `main` after this branch, writes `_driveApi`, which this branch turns into a cache derived from the authenticator's client. Keeping both makes injection silently ineffective.
- `mediaHttpClient()`, also newer, needs the `_allowSilentAuth` opt-in that moved into the mobile authenticator.
- This branch adds a whole `CFBundleURLTypes` key to the macOS Info.plist; `main` has since added one for Adobe Lightroom's OAuth callback. Two such keys in one plist is invalid and only the first survives, so one of the two callbacks would have broken with no build error.

#1105 also fixes a live bug this branch predates: Google Drive was offered as a media-store backend on Windows and Linux, where `google_sign_in` has no implementation and the connect flow could never succeed.

The branch is left in place as provenance for the port; nothing here is lost.

PR #475, opened 2026-07-04
fix(sync): publish post-adopt edits as base-less changesets, not a full base

ericgriffin:
Fixes the sync unreliability introduced by #450: since the deferred self-base, changes made after a Replace-adopt didn't appear on other devices until much later, unpredictably.

## Root cause

#450 deferred the post-adopt self-base publish until the device had a local change — but when that first edit arrived, the writer (with no cloud manifest of its own; the replace wiped the folder) published the **entire library as a fresh base**. Every peer held no cursor for that device, so each one had to download the whole base just to reach the one edit behind it. Until that full upload *and* each full download completed (iCloud propagates base parts lazily; the reader silently skips a peer whose parts are incomplete and retries next sync), **none** of the editor's changes were visible anywhere. Pre-#450 the same heavy exchange happened at restore time, where slowness was expected; #450 moved it to first-edit time, where it read as broken sync.

## Fix

**Base-less changesets while the adopt marker holds** (`ChangesetWriter.publish`): with the null-`baseSeq` marker set and no cloud base, publish the delta as a changeset against the adopted library's max HLC (recorded by `markAdoptedPendingBase` — now load-bearing), writing a manifest with a null `baseSeq`. `ChangesetReader.pull` already cold-starts base-less manifests by applying changesets from seq 1; the adopted content itself comes from the epoch peers' bases, which their logs retain. The watermark is safe by construction: `ensureSyncClockConfigured` seeds lazily from the max on-disk row HLC at the write choke point, so every post-adopt edit exceeds it.

**Compaction folds the log into a real base**: the count trip now measures from seq 0 for base-less logs, so after `compactionMaxChangesets` the deferred base finally publishes once, amortized, keeping the device a durable cold-start source.

**Lost-edit race closed** (`performSync`): `clearPendingRecords` / `clearAllSyncRecords` now run only when a publish was attempted. Previously, an edit landing between the skip decision and the end-of-sync cleanup had its pending row wiped — every following sync skipped again and the edit stayed unpublished until an unrelated later edit re-tripped the gate.

The skip gate deliberately stays pending/conflict/deletion-row based rather than HLC-based: rows merged from peers during pull sit above the device's own watermark, so an HLC gate would force passive viewers to publish echo changesets — recreating the very upload #450 eliminated.

## Verification

- New writer tests: first post-adopt publish is a changeset (no base parts uploaded), seq appends, clean no-op, deletion-only tombstone changeset, base-less log folds into a real base at the compaction trip.
- New reader test: cold-start from a base-less manifest applies changesets in order and advances the cursor (`baseSeqApplied` null).
- Epoch service tests updated: the first post-adopt edit publishes a base-less changeset carrying only the delta; a gate-blind race test proves a skipped sync leaves pending records intact.
- Full sync suite green (390), settings providers green (175), whole-project analyze + format clean.

### DIFF
--- a/lib/core/services/sync/changeset_log/changeset_writer.dart
+++ b/lib/core/services/sync/changeset_log/changeset_writer.dart
@@ -61,12 +61,22 @@ class ChangesetWriter {
     // recovers the seq counter (knownHeadSeq) so the new base never reuses a
     // number.
     final hasBase = ownManifest?.baseSeq != null;
+    // The post-adopt marker (a publish-state row with a null baseSeq): this
+    // device's library IS the adopted epoch the peers already published, so
+    // publishing its own base would redundantly re-upload the whole library --
+    // and every peer would have to re-download it just to reach the deltas
+    // behind it (the post-adopt "changes don't show up" unreliability). While
+    // the marker holds, publish changesets with NO base: a reader cold-starts
+    // a base-less manifest by applying changesets from seq 1 (the adopted
+    // content itself comes from the epoch peers' bases). Compaction
+    // eventually folds the log into a real base, clearing the marker.
+    final adoptedNoBase = !hasBase && state != null && state.baseSeq == null;
     final watermark = ownManifest?.publishedHlcHigh ?? state?.publishedHlcHigh;
 
     final newSeq = knownHeadSeq + 1;
     final now = DateTime.now().millisecondsSinceEpoch;
 
-    if (!hasBase) {
+    if (!hasBase && !adoptedNoBase) {
       // Stream the base to a temp file and slice-upload it, so a large library
       // is never materialized in RAM (#358 write side). Do NOT call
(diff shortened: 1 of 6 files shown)


ericgriffin:
Addressed the review finding about `markAdoptedPendingBase` receiving a possibly-null `maxRowHlc()` in b1cf46a76e9. Verified: the null case is real (empty adopted library, or rows that all predate HLC stamping), and a null watermark in the base-less mode would have routed through `exportChangeset(null)` -- the in-memory full-library export this PR exists to avoid.

Fixed in the writer rather than at the adopt site so ANY null-watermark marker is covered regardless of origin: `adoptedNoBase` now also requires a non-null watermark, falling back to the streamed base path. An empty adopted library no-ops there (zero rows, no deletions); a legacy no-HLC library publishes one bounded, streamed base. Regression test added: 'a null adopted watermark falls back to a streamed base publish'.

Issue #480, opened 2026-07-04
Large Shearwater logbook downloads over BLE time out and cannot resume; dives behind deleted records are never imported


john120283:
## Symptoms

- Downloading a large logbook from a Shearwater dive computer (Teric, and any device on the `shearwater_petrel` backend: Perdix, Petrel 2, Peregrine) over BLE often fails with a timeout partway through.
- A failed or cancelled download cannot resume: the next attempt re-downloads from the top and typically fails the same way, so owners of large logbooks can never complete an import.
- Additionally, dives that sit behind a deleted dive in the device's manifest can be silently skipped and, once a newer fingerprint is saved, are never imported at all.

## Root causes

All in the vendored libdivecomputer fork, `src/shearwater_petrel.c` (`shearwater_petrel_device_foreach`):

1. **Newest-first delivery breaks resume.** The profile-download loop walks the manifest records front to back, i.e. newest to oldest. Submersion persists the newest successfully-imported dive's fingerprint (`selectNewestFingerprint`) as the resume point. With newest-first delivery, a partial run saves the top-of-logbook fingerprint immediately, so the next run's manifest stop-check matches at the newest dive and every older dive is stranded.
2. **No per-dive resume.** A single failed per-dive download aborts the entire pass, and nothing marks where it stopped.
3. **Manifest truncation behind deleted records.** The manifest phase appends only `count * RECORD_SIZE` bytes per page, but the record walk advances past deleted (`0x5A23`) records without counting them. A deleted dive interspersed among not-yet-downloaded dives pushes the trailing valid record(s) past the appended prefix; those dives are never downloaded, with no error.
4. **Progress never reaches 100 percent on devices with deleted dives.** Deleted records are counted into the progress `maximum` but never credited to `current`.

## Fix

Branch `feat/shearwater-oldest-first`:

- Reverse only the profile-download loop so dives are delivered oldest to newest (fork patch `0003-shearwater-oldest-first.patch`). Because the loop stops on the first failed dive, a partial run always leaves a contiguous oldest-first prefix, which makes the newest imported fingerprint a correct high-water mark and gives working resume with no extra checkpoint bookkeeping. The manifest phase is unchanged (its newest-first walk is required by the fingerprint stop-check).
- Append all walked manifest bytes so records behind deleted dives are preserved, and exclude deleted records from the progress maximum (fork patch `0004-shearwater-manifest-deleted-records.patch`).
[...]

PR #634, opened 2026-07-19
fix(shearwater): restore newest-first downloads broken since build 114 (#621)

### LINKED ISSUE #621: DC import issues in build 116
Potential issue reported importing from Shearwater DCs on Android/Windows/? in build 116.

ericgriffin:
## Summary

Fixes #621: every Shearwater download has failed on real hardware since build 114 — on Android and Windows alike — with the download aborting "at the end of the transfer of the dive list" and zero dives imported. Reported against a Teric, a Perdix 2 AI, and a Petrel 3.

## Root cause

The submodule delta between build 113 (last known good) and build 114 is exactly the two libdivecomputer fork commits shipped for #480:

- `714a2e5` reversed the dive-download loop so dives were delivered oldest-first (to make the newest-imported fingerprint a valid resume point after a partial download).
- `7e8fd9a` made the manifest phase preserve records behind deleted (`0x5A23`) entries, so those dives were requested for the first time ever.

Both changed the request pattern sent to real Petrel-family firmware. No Shearwater client (Shearwater Cloud, Subsurface, upstream libdivecomputer) has ever requested dive payloads in that order, and PR #504 was validated only against the mocked transport in `test_shearwater_petrel_foreach.c` — never against hardware. `shearwater_common_download` treats any unexpected response as fatal (`DC_STATUS_PROTOCOL`), so the first request the device rejects — under oldest-first, the very first dive requested — aborts the entire pass. That yields the reported signature: manifest ("dive list") transfer completes, then immediate failure, nothing imported, on every attempt.

A secondary effect explains the "the computer suddenly wants to pair" reports: repeatedly aborted sessions poison the Android bond state (`onConnectionStateChange status=5`, `GATT_INSUFFICIENT_AUTHENTICATION` — the stale-bond case already documented in `BleIoStream.kt`). The BLE connect path itself is unchanged since build 113.

## Fix

**libdivecomputer fork** (`submersion-patches`, commit `45627c5`, pushed): revert `714a2e5` and `7e8fd9a`. `shearwater_petrel.c` is back to the upstream-proven newest-first behavior (only the Perdix 3 V2 signature changes remain vs. build 113). Patch mirrors `0003`/`0004` deleted; submodule pointer bumped.

**Native regression test**: `test_shearwater_petrel_foreach.c` rewritten to lock newest-first delivery as a hardware contract (verified red against the oldest-first driver, green after the revert; full native `ctest` suite passes 7/7). Two upstream limitations are deliberately restored and documented in-test rather than "fixed", because changing them alters the on-device request pattern: records pushed past the count-sized append prefix by deleted entries are not requested, and progress does not reach 100% on devices with deleted dives.

**App side — #480's resume goal is preserved, moved where it can't break the protocol**: the resume fingerprint now only advances after a download that ran to completion. `DiveComputerAdapter.setDownloadedDives` gains a `downloadComplete` flag (wired from `DownloadState.isComplete` in the wizard), and `_updateComputerAfterImport` skips `updateLastFingerprint` when the download was interrupted. A partial import still keeps its dives — the next download re-fetches from the top and duplicate detection skips them — but can no longer strand the never-delivered older dives. Import-phase cancellation still advances the fingerprint safely (the import loop is chronological, so the processed set is an oldest-prefix of a complete delivery).

## What #480 loses (follow-up)

Device-side resume for very large logbooks (the original #480 pain: a huge first download that times out restarts from the top each attempt) is intentionally regressed to the pre-114 state, which is at least the state where downloads work at all. A real re-fix — e.g. a per-record pre-download skip callback so already-imported dives are filtered without changing request order — needs hardware-validated design first. Same for the behind-deleted-records data loss.

## Testing

- Native: `ctest` 7/7 including the rewritten foreach regression test (15 assertions).
- Dart: full `test/features/import_wizard/` + `test/features/dive_computer/` suites (901 tests) pass; new regression test locks "interrupted download imports dives but keeps the old fingerprint".
- `flutter analyze` clean, `dart format` clean.

## Smoke test (requested)

On the Android device that currently fails: first **forget the dive computer in Android Bluetooth settings and power-cycle the DC** (repeated failed sessions from the broken builds leave stale bond keys — the `status=5` / unexpected pairing-prompt symptom), then download from a Teric/Perdix/Petrel. Expected: full download completes; a second download imports only new dives.


ericgriffin:
Closing to restart the investigation with corrected evidence: builds 114-116 with oldest-first delivery worked on macOS and iOS against real hardware, which contradicts this PR's premise that the firmware rejects the request order. The failure is platform-specific (Android bonding via the connect path, Windows still undiagnosed). A fresh fix will follow.

crcobb:
Testing on Windows version.
- Can connect and download dives from a Perdix AI.
- I downloaded 652 dives but at the end of the download I received a Download Failed message.  I was able to import all of the dives.  I'm investigating why about 99 of the dives didn't match existing dives.
- I then attempted to download again from the computer.  It started the download process again and I canceled after about 10 dives.  So either the error during the first download prevented the process from recognizing that dive had been downloaded or there is a problem in that process.  All the downloaded dives were identified as duplicates.
- I then tried downloading again and it restart downloading all dives.  The behavior in Subsurface if you cancel and import the downloaded dives, it recognizes that you have synced all the old dives on the computer.  Then next download it only downloads new dives or tells you, you have no new dives.  This is a nice option if somehow the software loses track of dives downloaded on a device you can quickly resync by downloading the new dives and pressing cancel.

If there is more info/logs I can provide please let me know.

DownloadFailure

crcobb:
Android build stills fails after downloading Perdix divelist.

Issue #645, opened 2026-07-19
Multiple dive-computer records can represent the same physical device


mikecarr:
## Problem

The same physical Shearwater Petrel appears as different saved dive computers on different devices because the app matches computers by exact Bluetooth identifier.

Current observations:

- Mac finds the known computer as `Petrel 3`.
- iPhone finds the known computer as `ssss`.
- iPad detected the Petrel under a third identifier and classified it as a new computer.
- The iPad log showed the current peripheral as `36A2CC68-5832-DF3E-EB1D-BD32304C4207`.
- Existing saved identifiers attempted by the iPad were `DB9A97A2-C3A6-7469-4864-3DDD13C0BD50` and `9337BCB9-6297-360B-D1FD-E262B2AA967A`; both failed to resolve.
- The iPad scan successfully matched the current peripheral as `Shearwater Petrel 3`.
- The import was canceled, so no third record was intentionally created during that test.

The records are likely aliases for one physical computer, not separate devices. Bluetooth identifiers can differ by platform or change after reinstall/reset, while the serial number should provide a stable identity.

## Expected result

When a scanned dive computer has a verified matching manufacturer/model/serial number, update or merge the existing computer record and its Bluetooth identifiers instead of creating another record. Historical dives should remain associated with the canonical computer record.

Do not delete either existing record until identity reconciliation is implemented or the records have been safely reviewed.

Related:
- #653 tracks the unrelated iCloud backup-folder bug.
- #654 tracks the separate dive-computer deletion foreign-key failure.
- #652 tracks stale iCloud sync state after database reset/reimport.
- PR #646 addresses the separate sync epoch/manifest protocol issue.

mikecarr:
[submersion-backup.log](https://github.com/user-attachments/files/30168086/submersion-backup.log)

mikecarr:
[submersion1.log](https://github.com/user-attachments/files/30168090/submersion1.log)

mikecarr:
## Confirmed cross-device BLE identifier issue

Further testing confirms the Petrel 3 can sync successfully, but the saved BLE identifier was device-specific and had been synced between devices.

- Existing iCloud-synced record used: `DB9A97A2-C3A6-7469-4864-3DDD13C0BD50`
- Fresh iPhone scan found the same Petrel 3 as: `9337BCB9-6297-360B-D1FD-E262B2AA967A`
- Stable device identity: serial `3101949313`
- Model: Petrel 3, libdivecomputer model 10
[...]

PR #646, opened 2026-07-20
Keep epoch fence strict and report skipped sync peers

mikecarr:
## Summary

Keep library-epoch filtering fail-closed after a restore/replacement, while making skipped stale peers visible to the user instead of reporting a misleadingly clean sync.

## Problem

Once a device accepts a library epoch, peer manifests with a different epoch—or no epoch at all—cannot safely be merged. An unstamped peer may still contain the replaced-away library, and its `updatedAt` can be refreshed by a heartbeat without proving that it observed or adopted the replacement.

The previous implementation attempted to use the manifest timestamp as an exception. That is unsafe because wall-clock time does not establish library-generation membership and is vulnerable to clock skew and heartbeat rewrites.

## Fix

- Remove the timestamp-based legacy exception.
- Keep the epoch fence strict for unstamped and differently stamped peers.
- Record skipped peer device IDs in `ChangesetReadResult` and `SyncResult`.
- Include a user-facing sync message explaining that those devices still need to adopt the current library.
- Remove `replacedAt` cutoff plumbing from all epoch-gate paths.
- Add regression coverage for current-epoch, stale-epoch, and unstamped peers, including reporting of skipped peers.

This keeps replaced-away data fenced while addressing the original silent-success/no-op behavior.

## Related issues

- #652 — iCloud sync can retain stale peer state after database reset or reimport; this is the incident/root-cause investigation.
- #645 — Petrel 3/dive-computer identity and deletion issues, including the original backup-path observation.

## Validation

- Dart formatting passes.
- `git diff --check` passes.
- The focused Flutter test is currently blocked by pre-existing generated Drift/database compilation errors on this branch, unrelated to this PR.


ericgriffin:
Here's Fable's analysis of this PR:

**Verdict**

The PR is pointing at a real problem, but the fix is unsound — it re-opens the exact data-leak the epoch protocol was built to prevent, and it's incomplete even on its own terms.

**The real issue underneath**

The failure mode the author hit is genuine: once a device accepts a library epoch, ChangesetReader silently skips every peer whose manifest is unstamped (epochId: null). The sync then reports success while having applied nothing — no error, no banner, nothing telling the user that two of their three devices are being ignored. That's a real reliability/UX gap in the epoch rollout story, and the author observed it on real hardware (it grew out of the multi-device Petrel 3 debugging in #645).

There's also an unanswered diagnostic question the PR skips over: why were the iPhone and Mac publishing unstamped manifests at all? The epoch gate is fail-closed — an epoch-aware device that sees the marker either halts for adoption or stamps everything it writes. So unstamped-post-epoch manifests can only come from (a) devices on old app builds that predate the protocol, or (b) devices that haven't seen the marker yet — plausible on iCloud, where file propagation lags. Either way, the peers are unaware of the epoch, and that's exactly what makes the fix dangerous.

**Why the fix is unsafe**

The PR's rule is: accept an unstamped peer manifest if manifest.updatedAt >= marker.replacedAt. Three problems, one fatal:

1. A timestamp can't tell library generations apart. The epoch fence exists for this scenario: you do a "Replace" restore, and a peer that still has the replaced-away library keeps publishing it. That peer is alive, so it keeps rewriting its manifest with fresh updatedAt values — every one of them after replacedAt. Under this PR, the freshly restored device merges the replaced-away library right back in. "Written after the replace" is not evidence of belonging to the new generation; for an epoch-unaware peer it's precisely the dangerous case. The protocol uses epoch IDs instead of timestamps for exactly this reason — the stamp is an affirmative statement "I have adopted the new library," which a timestamp can never be.
2. updatedAt isn't even a data-freshness signal. changeset_writer.dart:162-190 rewrites the manifest as a heartbeat — "contents unchanged, fresh updatedAt" — so peers don't read as dead devices for retirement/tombstone-GC purposes. A completely idle legacy peer carrying purely pre-epoch data will pass the cutoff after one heartbeat.
3. The comparison crosses wall clocks. updatedAt comes from the peer's clock, replacedAt from the restoring device's clock; skew flips the decision either way. And LibraryEpochMarker.fromJson defaults replacedAt to 0 when the field is missing (library_epoch.dart:60), which would make the filter fully open for every unstamped peer.
  
  **Why it's incomplete even on its own theory**

- The gate has three proceed sites; the PR threads the cutoff through two of them but not the _recoverUnreadableEpoch success path (sync_service.dart:695-697), so a device that re-established the backend from its local library would still silently skip legacy peers — the same symptom the PR claims to fix.
- The self-heal branch uses epochStore.lastAcceptedMarker?.replacedAt, which can be null while accepted is non-null (the epoch ID is also mirrored in _syncRepository) — filter silently stays strict there too. Safe, but inconsistent.
- The tests inject legacyEpochCutoffMs directly with now ± 1000, so nothing covers the gate actually wiring marker.replacedAt through, the replacedAt == 0 fallback, or the heartbeat-rewrite case.
- Mechanically, the PR is currently marked CONFLICTING with main.

**What a right-shaped fix would look like**

The legitimate scenario (peers with the same library that merely haven't stamped yet) is transient — once the marker propagates, those peers halt, adopt, and start stamping, and sync resumes. The durable bug is the silence. A safer fix would surface skipped stale-epoch peers in the SyncResult ("2 devices are still on the old library and need to adopt the replacement") rather than guessing from timestamps which unstamped peers are safe to merge. If auto-merging legacy peers is truly wanted, the gate should be something with actual provenance semantics (e.g., the peer's HLC watermark continuity against the adopted base), or an explicit user confirmation — not wall-clock arithmetic.

★ Insight ─────────────────────────────────────
This is a classic distributed-systems trap: substituting physical time for causal/generation identity. The epoch ID is a Lamport-style generation fence — membership is proven by having observed the replace event, not by when you last wrote. Any timestamp-based relaxation of such a fence reintroduces the split-brain it was built to prevent, because the nodes you most need to fence out (the ones that never saw the event) are exactly the ones that keep writing with fresh timestamps.
─────────────────────────────────────────────────

**Short version:** real symptom, correct observation that sync goes silently inert, but the patch trades a visible availability problem for an invisible data-integrity one. I'd not merge it as-is — the reader-side timestamp exception should be replaced with either surfacing the skip or a provenance-based check, and any fix needs to cover the third gate path.

mikecarr:
Thanks for the detailed review. I agree with the core criticism: the timestamp-based exception is not safe.

A peer that missed the replace marker can continue publishing the replaced library, and its heartbeat or manifest update would still have an `updatedAt` after `replacedAt`. That does not prove adoption of the new epoch, so this patch could reintroduce the data-mixing problem the epoch fence is intended to prevent.

The additional gaps you identified are valid as well:

- the unreadable-epoch recovery path does not carry the cutoff;
- a missing `replacedAt` can make the comparison unsafe;
- the tests do not cover the actual epoch-gate wiring or heartbeat behavior.

I am going to rework this PR rather than merge the timestamp exception. The safer direction is to keep the strict epoch fence and surface skipped unstamped/stale peers explicitly, so sync cannot report an apparently successful no-op without explaining that those devices still need to adopt the current library epoch. A future automatic merge would need causal/provenance evidence or explicit user confirmation, not wall-clock time.

mikecarr:
I pushed a revision in commit `9bafbb225`.

The timestamp-based legacy exception has been removed. The reader now keeps the epoch fence strict: once a device has accepted an epoch, unstamped or differently stamped peer manifests are not merged, regardless of `updatedAt` or heartbeat timing.

Instead, the pull result records the skipped peer device IDs, and `SyncResult` exposes them with a user-facing message explaining that those devices still need to adopt the current library. This keeps replaced-away data fenced while making the previous silent-success/no-op behavior visible.

The epoch-gate plumbing no longer carries `replacedAt` as a filtering cutoff, including the self-heal path. The epoch reader tests now verify that an unstamped peer is reported and remains skipped even if its manifest is recent.

This PR remains focused on the protocol symptom. The database-reset/reimport hypothesis and reset UX safeguards are tracked in #645.

Validation: formatting and `git diff --check` pass. The focused Flutter test is currently blocked by pre-existing generated Drift/database compile errors on this branch, unrelated to these changes.

mikecarr:
Small follow-up amendment pushed as `ec817025b`: the stale-epoch test now also verifies that the skipped peer ID is surfaced. The PR implementation remains unchanged.

mikecarr:
The PR title and description have been updated to reflect the narrower protocol scope. The incident hypothesis and reset/reimport investigation now live in #652; #645 remains focused on the dive-computer identity/deletion track.

ericgriffin:
Let's please remove .agent/CONTINUITY.md, but otherwise looks good to me - thanks.

Copilot now reviewing so may require some other small follow-ups.

mikecarr:
Addressed the two Copilot follow-ups in commit `73806547b`:

- The epoch regression test now rewrites two unstamped peer manifests with deliberately different timestamps (`0` and the current time) and verifies both are skipped and reported.
- Sync completion messages now use singular/plural grammar and combine skipped-peer and adopted-identity messages instead of hiding the adoption event.

`dart format` and `git diff --check` pass. The focused Flutter test remains blocked by the existing generated Drift/database compilation mismatch on this branch.

ericgriffin (approved):
👍

PR #676, opened 2026-07-21
feat(profile): stepped deco stop band on the dive profile chart

readme42:
Renders decompression stops on the dive profile as a stepped, translucent band spanning from the stop depth up to the surface, switchable between app-calculated and dive-computer values.

## What it does

- **Calculated source** quantizes the ceiling curve up to the next multiple of the diver's `decoStopIncrement` (3.0 m default).
- **Computer source** uses the dive computer's own stop depths raw, with no re-rounding, since some computers use non-3 m stop spacing.
- The band resolves its source *independently* of the ceiling line, so switching the ceiling to Computer does not silently drag the band with it.
- Selecting Computer on a dive with no computer ceiling falls back to calculated and reports the resolved source through `MetricSourceInfo`.

Controls mirror the existing ceiling pair: a per-dive legend toggle with a source selector, plus app-level visibility and source settings. The band is a fill-only region with no stroke, and the legend swatch renders the same translucent block rather than a line.

## Schema

Adds two `diver_settings` columns, **v132 to v133**:

| Column | Type | Default |
| ------ | ---- | ------- |
| `show_deco_stops_on_profile` | BOOL NOT NULL | true |
| `default_deco_stop_source` | INTEGER NOT NULL | 1 (calculated) |

Both use PRAGMA-guarded idempotent helpers called from the `onUpgrade` block and the `beforeOpen` backstop, and both keys are seeded in `_applyDiverSettingDefaults` so sync payloads predating the columns still import.

The migration was originally written as v130 and has been renumbered twice on rebase, to v133, as v130 (media enrichment HLC), v131 (service clock reconcile) and v132 (bottom-time backfill) landed on main in the meantime. `migration_v133_deco_stops_test.dart` now holds the exact-latest schema tripwire.

## Notable implementation detail

The fill is an `aboveBarData`, not a `belowBarData`. Depths are negated for the inverted axis, so the surface (y = 0) sits above the stop depth (y = -6) and the shaded region is the area above the bar. A below-bar fill cannot express this: fl_chart draws the below-bar area and then erases the entire above-line region to clean up cut-off overdraw, wiping exactly the area the band needs. A test pins this so it does not regress.

## Testing

Nine commits, each written test-first and independently reviewed. New coverage: quantization edge cases, source resolution in both directions, step-transition decimation, the fill and stroke configuration, chart layering, migration and legacy sync payload handling, and two tests proving the legend's source actually reaches `overlayComputerDecoData` at both call sites.

`dart format` clean, `flutter analyze` clean. The full `test/core/database/` suite (232 tests) passes against the renumbered v133 migration.

Coverage and review follow-up: a whole-branch review found that the diver's `decoStopIncrement` never reached the quantizer, so the band was always on a 3 m grid regardless of the setting; it is now threaded through all three `ProfileAnalysisService` construction sites and the isolate input, with a provider-level test that asserts the band lands on a 2 m and a 5 m grid. The migration tests were rebuilt against real upgrade, idempotency and `beforeOpen` backstop paths, replacing onCreate-only assertions that would have passed with the migration deleted. The deco stop row was added to the external tooltip used by the panel and fullscreen readouts. Legend and settings coverage gaps flagged by Codecov are also closed. The migration and settings assertions were mutation-checked by removing the production code and confirming the tests fail.

### Menu: 
image

### Dive overview

image

### Dive details
image

## Related

- #677 — Subsurface XML import discards per-sample `stopdepth`, which is why the Computer source falls back to Calculated on dives imported from that format.
- #678 — the ceiling line's shaded fill has the same below-bar defect this PR fixes for the band.


ericgriffin (approved):
👍

PR #682, opened 2026-07-21
Keep dive computer BLE identities device-local

mikecarr:
## Summary

Addresses #645, where the same physical dive computer can appear as separate records on different Apple devices.

CoreBluetooth identifiers are host-specific, so the synced bluetoothAddress must not be used as the computer identity. This PR:

- Keeps BLE addresses local to each device and excludes them from sync export/import.
- Prevents incoming records or conflict resolution from overwriting a local BLE address.
- After a newly discovered computer downloads successfully, matches it by manufacturer, model, and serial number and rebinds the current device-local address to the existing record.
- Creates a new record when the stable hardware identity is genuinely new.

This handles the Petrel 3/ssss scenario without conflating computers of the same model with different serial numbers. Existing duplicate records are not automatically merged by this PR.

## Validation

- flutter test test/features/import_wizard/data/adapters/dive_computer_adapter_test.dart
- flutter test test/core/services/sync/sync_serializer_fetch_record_test.dart
- dart analyze on all changed implementation and test files
- git diff --check


mikecarr:
Reason for this PR is because if I sync my Petrel on my iPhone, that device gets persisted. Now if I try and sync the same same device on my iPad it only sees it as a new device. I can withdraw this if I got it wrong

ericgriffin (approved):

PR #683, opened 2026-07-21
fix(dive-log): treat a gas change at the first sample as the initial gas (#679)

### LINKED ISSUE #679: Gas timeline strip shows a spurious sliver of the wrong gas at the start of imported dives
## Symptom

Every imported dive shows a very short segment of a different gas at the very start of the gas-usage strip, immediately followed by the real gas for the rest of the dive. In the reported case the sliver renders green (Nitrox, `gas_colors.dart:21`) ahead of the orange Air segment (`gas_colors.dart:18`).

Image

## Mechanism

`buildGasUsageSegments` in `lib/features/dive_log/data/services/gas_usage_segments_service.dart:74` synthesises an initial segment covering `0 .. firstSwitch.timestamp`, using the **lowest-`order` tank**, and only skips it when the first switch is at exactly `t == 0`:

```dart
final firstSwitch = inBoundsSwitches.first;
if (firstSwitch.timestamp > 0) {
  segments.add(GasUsageSegment(startSeconds: 0, endSeconds: firstSwitch.timestamp, gasMix: startingTank.gasMix, ...));
}
```

But dive computers do not sample at `t = 0`, and importers record the initial gas as a gas-change event at the **first sample**, not at zero. From `test/dives/001_short_deco_single_gas_switch.ssrf.xml`:

```
<sample time='0:10 min' ... />
<event time='0:10 min' type='25' flags='2' name='gaschange' cylinder='0' />
<event time='25:10 min' type='25' flags='1' name='gaschange' cylinder='1' o2='50.0%' />
```

The first event is at 10 s and declares the *starting* gas (note `flags='2'` versus `flags='1'` on the genuine mid-dive switch). Because 10 > 0, the code invents a 10-second segment before it and fills it from the lowest-`order` tank rather than from the gas that event names.

When the lowest-`order` tank happens to be the same mix as the first switch, `_mergeAdjacentSameGas` (`:108`) hides the artefact — which is why the committed fixtures do not show it. When they differ (a deco/stage cylinder sorting first), the sliver becomes visible as the wrong gas.

## Why it matters

- It misreports what was breathed at the start of the dive, which is gas-planning-relevant data, not only a cosmetic seam.
- `buildActiveTankIntervals` (`:167`) has the identical `inBounds.first.timestamp > 0` guard, so the same phantom interval is attributed to the wrong cylinder there. That map drives per-tank pressure interpolation, so the error can propagate beyond the strip.

## Suggested fix

Treat a gas-change at or before the first profile sample as the initial-gas declaration rather than a mid-dive switch: pass the first sample timestamp into both functions and suppress the synthetic initial segment when `firstSwitch.timestamp <= firstSampleSeconds`. Apply the same change in `buildActiveTankIntervals` so the two stay consistent.

Falling back to the first switch's own gas for the `0 .. firstSample` interval would also work and keeps the strip starting at zero; either is preferable to the current lowest-`order` guess.

## To confirm before fixing

I reproduced the mechanism from the code and the committed fixtures, but not the reporter's exact data: in fixtures 001 and 004 cylinder 0 is the back gas, so the merge masks it. Worth dumping `tanks[].order` and the first `gas_switches` row for one affected dive to confirm the lowest-`order` tank is the nitrox cylinder there.

## Suggested tests

- A first gas-change at the first sample (not at 0) produces no leading segment of a different gas, even when the lowest-`order` tank is a different mix.
- A genuine mid-dive switch still produces its own segment.
- `buildActiveTankIntervals` attributes the `0 .. firstSample` window to the tank named by the initial gas-change, not to the lowest-`order` tank.

Unrelated to #676; pre-existing.

readme42:
Fixes #679.

## Symptom

Imported dives render a short segment of the wrong gas at the very start of the gas-usage strip — green Nitrox ahead of the orange Air actually breathed.

## Root cause

`buildGasUsageSegments` treats any gas change at `t > 0` as a mid-dive switch and synthesises a leading segment from the **lowest-`order` tank**. But most dive computers do not sample at `t = 0`, and importers record the starting gas as a gas-change event at the **first sample**. Whenever the lowest-`order` tank holds a different mix than that event names, the invented segment renders as a sliver of a gas that was never breathed.

The condition is precisely **first sample != 0**. Dumping the parsed fixtures gives a clean controlled comparison:

| Fixture | Source | First sample | Initial declaration | Lowest-`order` tank | Sliver |
| --- | --- | --- | --- | --- | --- |
| 001 / 004 | Subsurface | `10` | cyl 0, Air | cyl 0, Air | hidden by merge |
| 002 | Subsurface | `10` | cyl 2, **Tx 16/48** | cyl 0, **EAN50** | **visible** |
| 003 | Subsurface | `10` | cyl 1, **Air** | cyl 0, **EAN40** | **visible** |
| 005 | Garmin FIT | **`0`** | `t=0` | Tx 19/34 | none |

003 is exactly the reported artefact: a green EAN40 sliver ahead of orange Air. Garmin samples at zero, so its declaration lands at `t = 0`, the original `> 0` guard is already false, and the FIT path was never affected.

Tank ordering does not cause the bug — it only decides whether the phantom segment is *hidden* (001/004, where `_mergeAdjacentSameGas` absorbs it) or *wrong-coloured* (002/003).

The screenshot in the issue corroborates this: the sliver spans exactly from `x = 0` to where the depth trace begins, and the orange gas starts at the same instant as the first sample.

## Why it surfaced only recently

The flaw is original to the gas strip (`36692d5cd`) — the file is otherwise byte-identical to that commit — but it was **unreachable**. `gas_switches` was empty for these dives, so the function took its `if (inBoundsSwitches.isEmpty)` early return and produced one correct full-dive segment. `cfcd956eb` ("persist gas switches on multi-gas dive-computer dives") started populating the table, and the buggy branch woke up.

Note `cfcd956eb` got the convention right for the dive-computer download path: `resolveGasSwitches` explicitly drops the baseline ("the first observed gas is the baseline... not a switch"). The file-import parsers never adopted it. That split convention is the underlying defect.

## Fix

Treat a switch at or before the first profile sample as the initial-gas declaration: it owns the window from `t = 0`, and no leading segment is invented. When several land in that window the last one wins. The same rule is applied in `buildActiveTankIntervals`, which drives per-tank pressure interpolation, so the two stay consistent.

`firstSampleSeconds` defaults to `0`, so behaviour is unchanged for callers without profile data — and provably identical on the FIT path, where `timestamp <= 0` is the old `> 0` guard inverted. Threaded through all call sites plus `synthesizeEstimatedTankPressures`.

Because the strip's axis starts at `viewMin = 0`, letting the declaration own `0 .. firstSample` keeps the strip starting at zero with no gap, rather than merely suppressing the sliver. The dive does start at `t = 0` — the computer simply has not sampled yet — so that window is real dive time and belongs to the gas carried in from the surface.

## Tests

Three behavioural tests, each verified to fail against the old behaviour (spurious extra segment; interval starting at `10` instead of `0`) and pass with the fix:

- a gas change at the first sample produces no leading segment of a different gas, even when the lowest-`order` tank is a different mix
- a genuine mid-dive switch still gets its own segment
- `buildActiveTankIntervals` attributes the `0 .. firstSample` window to the tank named by the declaration

Plus generality coverage, since the rule keys off each dive's own first sample rather than the 10s interval the Subsurface fixtures happen to use: first samples of 0, 1, 2, 4, 5, 10, 20, 30 and 60 seconds (Shearwater 1s/2s, Suunto 4s/5s, older units 20s/30s), and a boundary case confirming a switch one second past the first sample is still a real mid-dive switch.

`flutter analyze` clean; `dart format` applied. Full suite: 13571 pass, 27 skipped, 4 fail — three of those four fail identically on `main` at 3b6cc74d9 (pre-existing, unrelated: backup, equipment, media), and the fourth (`setup_apply_service`) passes in isolation on this branch and at base, so it is order-dependent flake in a feature this change does not touch.

## Scope and follow-ups (deliberately not in this PR)

This is a read-side fix. Nothing persisted changes, so no migration or re-import is needed — all consumers of the estimated-pressure path are chart widgets, and `synthesizeEstimatedTankPressures` is pure and never persists.

Known related items left alone:

1. `gas_analysis_service._getActiveTankAtTimestamp` has its own independent pre-first-switch fallback (`role == backGas`, else `tanks.first`) and mis-attributes the same `0 .. firstSample` window, feeding SAC segmentation. The error is ~10s of a ~40min dive, so numerically negligible, but it is the same root cause outside the display path.
2. Having the Subsurface parser stop emitting the initial declaration as a switch row — matching `resolveGasSwitches` — is the deeper consistency fix, but it only helps on re-import. The service-level fix is what repairs already-imported dives.
3. The depth trace itself still starts at the first sample while the axis starts at `0`, leaving a few pixels of empty space at the left edge. That is a separate rendering decision from gas attribution.

### DIFF
--- a/lib/features/dive_log/data/services/estimated_tank_pressure_synthesizer.dart
+++ b/lib/features/dive_log/data/services/estimated_tank_pressure_synthesizer.dart
@@ -22,6 +22,7 @@ EstimatedTankPressures synthesizeEstimatedTankPressures({
   required List<DiveTank> tanks,
   required List<GasSwitchWithTank> gasSwitches,
   required int diveDurationSeconds,
+  int firstSampleSeconds = 0,
 }) {
   final pressures = <String, List<TankPressurePoint>>{...existing};
   final estimated = <String>{};
@@ -33,6 +34,7 @@ EstimatedTankPressures synthesizeEstimatedTankPressures({
     tanks: tanks,
     gasSwitches: gasSwitches,
     diveDurationSeconds: diveDurationSeconds,
+    firstSampleSeconds: firstSampleSeconds,
   );
 
   for (final tank in tanks) {
--- a/lib/features/dive_log/data/services/gas_usage_segments_service.dart
+++ b/lib/features/dive_log/data/services/gas_usage_segments_service.dart
@@ -28,12 +28,20 @@ class GasUsageSegment {
 /// Algorithm:
 /// 1. Pick the starting tank (lowest [DiveTank.order], or the first tank).
 /// 2. Sort gas switches by timestamp and clamp to dive bounds.
-/// 3. The initial segment runs from t=0 to the first switch's timestamp
-///    (skipped if the first switch is exactly at t=0).
-/// 4. Each subsequent segment runs from one switch to the next, ending at
+/// 3. A switch at or before [firstSampleSeconds] is the dive's initial-gas
+///    declaration rather than a mid-dive switch, and owns the window from t=0.
+/// 4. Otherwise the initial segment runs from t=0 to the first switch, filled
+///    from the starting tank.
+/// 5. Each subsequent segment runs from one switch to the next, ending at
 ///    [diveDurationSeconds] for the final switch.
-/// 5. Adjacent segments with identical gas mixes are merged so a switch back
+/// 6. Adjacent segments with identical gas mixes are merged so a switch back
 ///    to the same gas does not produce a visible seam.
+///
+/// [firstSampleSeconds] is the timestamp of the dive's first profile sample.
+/// Dive computers do not sample at t=0, and importers record the starting gas
+/// as a gas-change event at that first sample; without it a phantom leading
+/// segment is invented from the lowest-order tank, which renders a sliver of
+/// the wrong gas whenever that tank holds a different mix (issue #679).
 ///
 /// Returns an empty list when there are no tanks or the dive has no
 /// duration — the caller should hide the strip in that case.
@@ -41,6 +49,7 @@ List<GasUsageSegment> buildGasUsageSegments({
   required List<DiveTank> tanks,
   required List<GasSwitchWithTank> gasSwitches,
   required int diveDurationSeconds,
+  int firstSampleSeconds = 0,
 }) {
   if (tanks.isEmpty || diveDurationSeconds <= 0) {
     return const <GasUsageSegment>[];
@@ -70,12 +79,16 @@ List<GasUsageSegment> buildGasUsageSegments({
 
   final segments = <GasUsageSegment>[];
 
-  final firstSwitch = inBoundsSwitches.first;
-  if (firstSwitch.timestamp > 0) {
+  final initialIndex = _initialGasDeclarationIndex(
+    inBoundsSwitches,
+    firstSampleSeconds,
+  );
+
+  if (initialIndex == null) {
     segments.add(
       GasUsageSegment(
         startSeconds: 0,
-        endSeconds: firstSwitch.timestamp,
+        endSeconds: inBoundsSwitches.first.timestamp,
         gasMix: startingTank.gasMix,
         label: startingTank.gasMix.name,
         tankName: startingTank.name,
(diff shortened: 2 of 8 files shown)


ericgriffin (approved):
👍

PR #698, opened 2026-07-23
fix(media): hydrate media taken_at as wall-clock-UTC

ericgriffin:
## Problem

Photos and videos added via **Add photo or video → Files** (folder pick) were all pinned to the first dive-profile point — surface depth, e.g. "2ft" — with the wrong time/depth association. The **gallery scan** path was unaffected.

This is distinct from #660, which fixed timestamp *extraction* on desktop. Extraction is correct here: the stored `media.taken_at` values match the source metadata (JPG EXIF and GoPro `mvhd`). The skew is introduced at **enrichment** time.

## Root cause

`MediaRepository._mapRowToMediaItem` hydrated `taken_at` with `DateTime.fromMillisecondsSinceEpoch(ms)` **without `isUtc: true`**, producing a local `DateTime`.

`taken_at` is persisted as **wall-clock-UTC** millis — the write path stores the epoch ms of an already-UTC `DateTime`, and the dive side reads `entry_time` with `isUtc: true` (`dive_repository_impl.dart:2237`). On a host with a non-zero UTC offset, the local hydration shows shifted digits (e.g. `11:47 UTC` → `06:47` on US-Eastern). `EnrichmentService.calculateEnrichment` then calls `toWallClockUtc()`, which reinterprets those shifted digits as UTC. The photo time is displaced by the host offset relative to the (correct, UTC) dive start, yielding a negative elapsed time so every item resolves to the first profile point.

Observed in a real DB: 11 enrichment rows all had `elapsed_seconds ≈ −16692` (−5h = host December offset) and `depth_meters = 0.4572` (≈ "2ft"), while the `taken_at` values themselves were correct.

## Fix

Hydrate `taken_at` with `isUtc: true` so downstream normalisation is a no-op.

- **Safe for `asset_resolution_service`**: it compares timestamps with `.difference()`/`.subtract()`, which operate on absolute instants and ignore the `isUtc` flag.
- **Corrects the photo viewer**: `DateFormat` now reads UTC components, showing the true wall-clock capture time instead of a host-offset-shifted one.

## Tests

- New repository round-trip guard asserting `fetched.takenAt.isUtc` (fails before the fix).
- Full media suite (725) + analyzer green.

## Notes

- Items imported *before* this fix keep their stale enrichment rows (the backfill only enriches rows that are missing); unlink + re-add to correct them.
- Local-file video thumbnails still render a movie-icon placeholder (no poster frame generated) — a separate, pre-existing gap; depth badges on videos are corrected by this change.
- Sibling `media_repository.dart:698` (GPS tuple) also omits `isUtc: true`; left unchanged as its use is instant-based, but worth a follow-up audit.

### DIFF
--- a/lib/features/media/data/repositories/media_repository.dart
+++ b/lib/features/media/data/repositories/media_repository.dart
@@ -1066,8 +1066,9 @@ class MediaRepository {
       longitude: row.longitude,
       // taken_at is stored as wall-clock-UTC millis (see the write path and the
       // dive side, which reads entry_time with isUtc: true). Hydrate it as UTC
-      // so downstream normalisation (EnrichmentService.toWallClockUtc) is a
-      // no-op instead of shifting the photo time by the host's UTC offset.
+      // so downstream normalisation (TripMediaScanner.toWallClockUtc, invoked by
+      // EnrichmentService.calculateEnrichment) is a no-op instead of shifting the
+      // photo time by the host's UTC offset.
       takenAt: row.takenAt != null
           ? DateTime.fromMillisecondsSinceEpoch(row.takenAt!, isUtc: true)
           : _defaultTakenAt(row.id),
--- a/lib/features/media/data/repositories/media_repository.dart
+++ b/lib/features/media/data/repositories/media_repository.dart
@@ -1064,8 +1064,12 @@ class MediaRepository {
       mediaType: _parseMediaType(row.fileType),
       latitude: row.latitude,
       longitude: row.longitude,
+      // taken_at is stored as wall-clock-UTC millis (see the write path and the
+      // dive side, which reads entry_time with isUtc: true). Hydrate it as UTC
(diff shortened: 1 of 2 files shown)


ericgriffin:
Thanks for the review. All three points addressed:

1. **`toWallClockUtc` attribution (repo comment + test):** Corrected in both places. The comments now reference `TripMediaScanner.toWallClockUtc` (the static conversion at `trip_media_scanner.dart:385`), invoked inside `EnrichmentService.calculateEnrichment` — no longer implying a non-existent `EnrichmentService.toWallClockUtc`.

2. **Orphan-prevention docs scope creep:** Removed from this PR. Those two files had already merged to `main` via the orphan-prevention work; they only appeared here because the branch's merge-base predated them. Rebasing onto latest `main` dropped the two doc commits as already-applied, so the branch is now purely the timestamp fix (`media_repository.dart` + the regression test).

The branch was force-pushed after rebase. Diff is now the two intended files only; media suite green locally.

PR #744, opened 2026-07-27
fix(dive-log): keep large dive numbers inside the list-tile badge

etlami:
## Problem

On both macOS and iOS, the dive number badge in the dive list tiles doesn't fit once the number gets large. The badge renders a fixed 12px bold `#<n>` in a fixed-size slot:

- detailed mode: a 40x40 `CircleAvatar`
- compact / dense mode: a 36px-wide leading slot

At four or five digits (e.g. **#2845**) the text no longer fits, wraps to two lines and overflows the circle. Reported on the dashboard Recent dives list, but it affects the Dives tab too since both reuse these tiles.

## Fix

Wrap the badge text in `FittedBox(fit: BoxFit.scaleDown)` with `maxLines: 1` in all three tile variants (`dive_list_page.dart` `DiveListTile`, `compact_dive_list_tile.dart`, `dense_dive_list_tile.dart`). Long numbers scale down to fit on a single line; short numbers keep the original 12px size (`scaleDown` never enlarges).

## Tests

Added a regression test to each tile's test file: rendering a 5-digit dive number keeps the badge on one line (`maxLines == 1`), wrapped in a `FittedBox`, with no layout exception.

Verified locally against Flutter 3.44.8 / Dart 3.12.2: `flutter test` on the three tile tests passes (41 tests), `flutter analyze` clean, `dart format` clean.

## Note

The same `#<n>` marker is also drawn on map pins (`dive_map_content.dart`, `dive_activity_map_page.dart`) and in `dive_profile_panel.dart`. Those are a different visual context and out of scope here; can follow up if they show the same clipping.

### DIFF
--- a/lib/features/dive_log/presentation/pages/dive_list_page.dart
+++ b/lib/features/dive_log/presentation/pages/dive_list_page.dart
@@ -839,12 +839,21 @@ class DiveListTile extends ConsumerWidget {
                             )
                           : CircleAvatar(
                               backgroundColor: colorScheme.primaryContainer,
-                              child: Text(
-                                '#$diveNumber',
-                                style: TextStyle(
-                                  color: colorScheme.onPrimaryContainer,
-                                  fontWeight: FontWeight.bold,
-                                  fontSize: 12,
+                              child: Padding(
+                                padding: const EdgeInsets.symmetric(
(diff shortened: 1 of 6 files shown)


ericgriffin (approved):
👍

PR #748, opened 2026-07-27
fix(deco): guard against non-bar surface pressure poisoning Deco Status / Tissue Loading

etlami:
## Problem

On some dives the **Deco Status** and **Tissue Loading** panels show impossible values for an ordinary shallow dive:

- NDL `998:59`, GF99 `0%`, SurfGF `0%`
- Tissue Loading: Load `30817%`, N2 `998.51 bar`

## Cause

Single upstream cause: the dive's stored **`surfacePressure` is in millibar/hectopascal magnitude (~1013-1264) but the field is typed and consumed as bar**. The deco engine and the panels are correct — they faithfully render a model fed a surface pressure ~1000x too large:

- compartments seed surface-saturated at `(surfaceP - 0.0627)·fN2` ≈ **998.5 bar** with surfaceP ≈ 1264 → the reported N2 `998.51 bar`
- `percentLoading = pN2 / surfaceMValue * 100` ≈ 998.5 / 3.24 ≈ **30817%**
- ambient pressure ≈ 1265 bar ≫ tissue tension → gradient factors clamp to **0%**
- tissues never exceed the (enormous) surface ceiling → NDL saturates at the cap `999*60 - 1` s, printed as **`998:59`** (an "unlimited NDL" sentinel shown literally)

Some import/legacy path stored a mbar/hPa value in this bar-typed field. The two active importers actually convert correctly (`UddfFullImportService` does `pascal / 100000`; the Subsurface parser strips the `bar` suffix), so the corrupt value is most likely legacy data or an unaudited path (MacDive/CSV/DL7/native download, or a non-conformant file), not those importers.

## Fix

Sanitize/repair the surface pressure at the source so every consumer — the deco panels **and** the display paths (the detail page rendered a stored `1264` as `1264000 mbar`; the edit form round-tripped it) — sees a clean value:

- Extract the clamp into a shared `normalizeSurfacePressureBar()` in `core`: convert an obvious mbar/hPa value (`> 100`) to bar, then reject anything still outside the physically-possible surface range (~0.5-1.1 bar).
- Normalize at the `Dive` load/persist boundary in `DiveRepositoryImpl` (both row→entity mappers and both write companions), so reads and writes are clean.
- Add a **v137 migration** that repairs stored `dives.surface_pressure` rows the same way. Since no real surface pressure is ever out of range, only corrupt rows are touched.
- Keep the `DiveEnvironment.forConditions` guard as defense-in-depth, since the planner/deco-calculator feed it user-entered values that never pass the DB boundary.

## Tests

- Unit test for `normalizeSurfacePressureBar` (mbar/hPa conversion, implausible-value rejection, boundary values).
- v137 migration test: a stored `1013` becomes `1.013`, an implausible `1264` becomes `NULL`, a valid `1.02` is untouched.
- `dive_environment_test` still covers `forConditions` behaviour; the v136 exact-version tripwire is relaxed per its own instructions.

Verified locally against Flutter 3.44.8 / Dart 3.12.2: `test/core/deco/`, `test/core/database/`, `test/core/utils/`, and the dive-repository suites pass; `flutter analyze` clean; `dart format` clean.

## Credit

Thanks to @readme42 for the review that corrected the original importer misattribution and prompted moving the fix from read-time to the source.

### DIFF
--- a/lib/core/database/database.dart
+++ b/lib/core/database/database.dart
@@ -2846,7 +2846,7 @@ class AppDatabase extends _$AppDatabase {
 
   /// The current schema version as a static constant so that pre-open checks
   /// (e.g. version-mismatch guard) can reference it without an instance.
-  static const int currentSchemaVersion = 136;
+  static const int currentSchemaVersion = 137;
 
   /// Every schema version that has a migration block in onUpgrade.
   /// Used to calculate progress step counts. When adding a new migration,
@@ -3010,6 +3010,9 @@ class AppDatabase extends _$AppDatabase {
     // Phase A). Renumbered from v130 as main advanced past it at merge time.
     134,
     136,
+    // v137: repair surface_pressure rows stored in millibar/hectopascal
+    // (used verbatim as bar, they poisoned the deco/tissue panels).
+    137,
   ];
 
   /// Idempotent DDL for the v106 connector-suggestion columns (Lightroom
@@ -3196,6 +3199,34 @@ class AppDatabase extends _$AppDatabase {
   /// for minimal old-schema fixtures. onUpgrade only -- never beforeOpen -- so
   /// a dive a user deliberately left with bottom_time == runtime (and no
   /// profile to prove otherwise) is never re-touched.
+  /// Repairs `dives.surface_pressure` rows stored in the wrong unit.
+  ///
+  /// The column is bar, but some import/legacy paths wrote millibar/
+  /// hectopascal values (e.g. 1013) into it. Used verbatim, a ~1000x-too-large
+  /// surface pressure poisons the decompression model and shows absurd figures
+  /// on the dive detail page. This mirrors [normalizeSurfacePressureBar]:
+  /// convert an obvious mbar/hPa value to bar, then null out anything still
+  /// outside the physically-possible surface range (~0.5-1.1 bar). Since no
+  /// real surface pressure is ever out of that range, only corrupt rows are
+  /// touched. PRAGMA-guarded like every other migration helper.
+  Future<void> _repairSurfacePressure() async {
+    final cols = await customSelect("PRAGMA table_info('dives')").get();
+    final colNames = cols.map((c) => c.read<String>('name')).toSet();
+    if (!colNames.contains('surface_pressure')) return;
+
+    // Obvious mbar/hPa magnitude -> bar.
+    await customStatement(
+      'UPDATE dives SET surface_pressure = surface_pressure / 1000.0 '
+      'WHERE surface_pressure IS NOT NULL AND surface_pressure > 100',
+    );
+    // Anything still physically impossible -> NULL (fall back to standard).
+    await customStatement(
+      'UPDATE dives SET surface_pressure = NULL '
+      'WHERE surface_pressure IS NOT NULL '
+      'AND (surface_pressure < 0.5 OR surface_pressure > 1.1)',
+    );
+  }
+
   Future<void> _backfillBottomTimeFromProfile() async {
     // PRAGMA-guarded like every other migration helper: minimal old-schema
     // test fixtures (and any DB that reached this block through a guarded
@@ -7063,6 +7094,13 @@ class AppDatabase extends _$AppDatabase {
           await _assertMediaStoresLastSweepColumn();
         }
         if (from < 136) await reportProgress();
+        // v137: repair surface_pressure values stored in millibar/hectopascal
+        // (e.g. 1013) in this bar-typed column, which poisoned the deco/tissue
+        // panels and displayed as absurd figures elsewhere.
+        if (from < 137) {
+          await _repairSurfacePressure();
+        }
+        if (from < 137) await reportProgress();
       },
       beforeOpen: (details) async {
         // Enable foreign keys
--- a/lib/core/deco/entities/dive_environment.dart
+++ b/lib/core/deco/entities/dive_environment.dart
@@ -40,14 +40,16 @@ class DiveEnvironment extends Equatable {
   ///
   /// An explicit [surfacePressureBar] wins over [altitudeMeters]. A null
   /// altitude keeps the legacy 1.0 bar surface so dives without altitude
-  /// data are unchanged.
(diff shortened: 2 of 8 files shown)


readme42:
Nice catch on this, and the write-up is genuinely excellent — the deco-model explanation is exactly right, and the guard is clean and well-tested. You correctly identified and neutralized the poisoning, which is the hard part.

A few things I turned up while reading around it — this is my own read as one contributor, not a call on whether it merges:

**The two importers in the "Cause" section actually convert correctly — I couldn't reproduce the bug from either.**
- UDDF converts Pa → bar (`pascal / 100000`, both the standard `atmosphericpressure` key and Shearwater's `surfacepressure`) at `uddf_full_import_service.dart:1218`, and has since well before the Feb 2026 refactor.
- Subsurface strips the `bar` suffix via `_parseDouble`, and only started importing this field in April 2026, so there's no window where it wrote raw values.

So I think the corrupt value is more likely legacy data or an unaudited path (MacDive/CSV/DL7/native download, or a non-conformant file) than those two. If you have a dive that reproduces it, I'd be happy to trace the real origin.

**The other thing I noticed:** the guard sanitizes at read time inside `DiveEnvironment.forConditions`, but the bad value stays in the DB, so other consumers still read it raw — the detail page shows a stored `1264` as `1264000 mbar` (`dive_detail_page.dart:3028`), and the edit form pre-fills/round-trips it.

**What I'd lean toward:** fixing it at the source instead — a small migration to repair the stored rows, plus a single normalization where a `Dive` is loaded/persisted. That would cover the deco panels *and* the display paths in one place, and since it'd land in the same release, users would be just as protected on day one as with the read-time guard — just more completely. It's low-risk too: no real surface pressure is ever outside ~0.5–1.06 bar, so a migration that only touches out-of-range rows can't really harm good data.

Happy to help with the migration/boundary version if that's useful, or to hand over what I dug up — whatever works best for you. Either way, really nice work.

etlami:
You're right on both points.

The active UDDF path (`UddfFullImportService`) converts `pascal / 100000`, and the Subsurface parser strips the `bar` suffix — neither writes a raw value. I've corrected the Cause section; the corrupt value is legacy/unaudited-path data, not those importers.

On the guard being read-time only: correct, the stored value survived — the detail page showed `1264000 mbar` and the edit form round-tripped it. Reworked as suggested:

- shared `normalizeSurfacePressureBar()` in `core`;
- normalization at the `Dive` load/persist boundary in `DiveRepositoryImpl` (both row->entity mappers and both write companions);
- a v137 migration repairing stored `dives.surface_pressure` rows (mbar/hPa -> bar, then null out anything still outside ~0.5-1.1 bar).

Kept the `DiveEnvironment.forConditions` guard as defense-in-depth for planner/deco-calculator inputs that don't pass the DB boundary.

If useful, I can try to pull a reproducing dive to trace the real origin.

readme42:
Thank you, looks promising!

> If useful, I can try to pull a reproducing dive to trace the real origin.

Yes, that would be great. Let's see where it comes from and/or if its already corrupt data. Then we might want to inform the user during import as well.

readme42:
@ericgriffin Would a "database_migration" label make sense to visualize PRs which easily create merge conflicts? It could be autolabeld by Github. 
Or is the whole merge process automated anyways?

etlami:
Traced it — the corrupt value comes from my own Petrel sync bridge, not a Submersion importer.

The bridge injects dives directly (bypassing the importers). It reads libdivecomputer's `<atmospheric>` field from dctool — already in bar, e.g. `1.00300` — and multiplied it by 1000 before injecting, so 1.003 bar was stored as 1003.

Why I got the unit wrong: Submersion presents surface pressure as **mbar** in the UI — the detail row renders `surfacePressure * 1000` as "mbar", and the edit form prefills ×1000 / saves ÷1000 — but the DB column is actually **bar**; the UI just does the ×1000/÷1000 for display. I wrote the injector to match what the app shows, so I stored mbar. Since dctool already gives bar, the fix on my end is to inject it as-is. (Might be worth a thought whether the mbar-labelled UI over a bar column invites this — but that's minor.)

So Submersion's own importers are fine, as you said. The v137 migration + load/persist normalization still make sense — they repair the already-corrupted rows and guard against any other non-conformant source like this one.

readme42:
Really nice detective work running this all the way down — tracing it to the Petrel bridge and pinning the exact `×1000` is exactly the kind of root-causing that saves everyone downstream. Thank you for being so transparent about it.

This is just my personal opinion as one contributor, but given where it landed: since the corrupt values come from the bridge injecting mbar into a bar column (and Submersion's own importers are all correct), I'd lean toward fixing it purely in the adapter rather than carrying anything in Submersion for it. My reasoning:

- The data was never produced by Submersion, so a migration in the shared schema history would really just be cleaning up one external tool's output — that feels like it belongs on the bridge side (inject bar directly + a one-off cleanup of the already-affected rows).
- I'm also a little wary of the read/write normalization as a permanent guard: silently correcting a physically-impossible value would have *hidden* this exact bug. The only reason it surfaced and got fixed at the source is that Submersion didn't quietly launder it — the absurd panel values are what made it visible. A guard that auto-fixes impossible input tends to suppress the signal that something upstream is broken.

So my vote would be to fix it in the bridge and not merge the migration/normalization here — but that's genuinely just my take, and I'd defer to you and the maintainers on it.

One thing from this that I do think is worth a separate look (unrelated to your PR): the detail page and edit form present surface pressure as **mbar** while the column is actually **bar** — which is exactly what led the bridge astray. Might be worth making that unit contract clearer somewhere so the next integration doesn't hit the same trap. Really appreciate you digging into this.

etlami:
Agreed on both points — the "silent guard would have hidden this exact bug" argument in particular is convincing. Since the corrupt values come solely from my external bridge (now fixed to inject bar) and Submersion's own importers are all correct, I'll close this rather than carry a migration/normalization in the shared schema for one tool's output. I'll clean up the already-affected rows on the bridge side.

Thanks for the thorough back-and-forth. I'll open a separate small item for the mbar-vs-bar unit contract on the detail/edit surfaces, since that's what led the integration astray.

Issue #757, opened 2026-07-27
Buddy filter does not work


schtibal:
**Describe the bug**
Using the dive list filter feature, filtering by buddy name doesn't work

**To Reproduce**
1. Go to 'Dives'
2. Click on 'Filter Dives'
3. Enter a known buddy's name
4. Click 'Apply Filters'

**Expected behavior**
The dive list should show all dives that include the buddy specified. Instead, no dives match.

**Platform details:**
- OS: Windows
- Version 1.7.0.117

**Additional context**
The Buddy list is a free text field, compared to the Dive Site, which is a selection list. Having a dropdown of buddies is impractical. Ideally it is a free text field that includes an auto-filtering drop down, exactly like the buddy entry screen when editing a dive

PR #769, opened 2026-07-30
fix - buddy filtering in dive logs 

jamespet77:
## Summary

Fixes an issue where buddy filter was not working correctly.  #757 
## Changes

- Implemented `buddyNameFilter` with autocomplete support in `DiveFilterSheet`, allowing user-friendly suggestions for buddy names.
- Extended backend filtering logic to match structured buddy data from `dive_buddies` junction table.
- Updated `DiveSummary` to include `buddyNames` for efficient summary-level filtering.
- Enhanced `DiveRepository` query to retrieve and map structured buddy data for dives.
- Added widget and domain tests to validate buddy filtering behavior and autocomplete interactions.
-

## Test Plan

- [x] `flutter test` passes
- [x] Manual testing on:


ericgriffin:
Thanks for tackling this. There are two changes I'd like to see before merge, plus a few smaller notes.

## Requested changes

### 1. Revert the `apply()` genericization in `dive_filter_state.dart`

The signature changed from `List<Dive> apply(List<Dive>)` to `List<T> apply<T>(List<T>)`, but the method immediately throws `ArgumentError` unless `T` is `Dive` or `dynamic`, and throws per-element for anything that isn't a `Dive`. So it accepts exactly what it accepted before — except every field access now goes through `as dynamic` (`dive.equipment as List<dynamic>`, `dive.bottomTime as Duration?`, etc.). That means the analyzer can no longer catch a renamed or mistyped field across ~10 filter axes; those would surface as runtime `NoSuchMethodError`s instead.

Since the guard guarantees the element is a `Dive`, the typed signature works as-is:

```dart
List<Dive> apply(List<Dive> dives) {
  return dives.where((dive) {
    ...
```

All the new buddy-filter logic (the comma split, legacy + junction matching) works unchanged with the typed version. The `throwsArgumentError` test can be deleted along with it. If the intent was to eventually filter `DiveSummary` in memory, that would be better as a separate typed method when it's actually needed — as written, `T == DiveSummary` throws.

### 2. `DiveSummary.buddyNames` is currently dead code with a per-page query cost

Nothing reads the new field — no filtering or display code in this PR consumes it, and it's a new field so no existing code can. But both summary call sites now run the extra `_buddyNamesForDives` batch query on every dive-list page load. Suggest dropping the field, the mapper changes, and `_buddyNamesForDives` from this PR; the SQL-level filtering already handles buddy matching for the list. (If summary-level filtering is planned, it can come back together with its consumer.)

Related: if it stays, note that `_mapSummaryRows` reads `row.readNullable<String>('buddy')`, but the summary SELECT doesn't include `d.buddy` — Drift returns `null` for a missing column rather than throwing, so the legacy-scalar branch silently never fires. A repository-level test asserting `buddyNames` contents would catch this.

## Smaller notes (non-blocking)

- **`LayoutBuilder` in `dive_filter_sheet.dart`** is used only to get a builder scope for `ref.watch(allBuddiesProvider)` and ignores its constraints. This is a `ConsumerState`, so `ref.watch` works directly in `build` — the wrapper can go.
- **The `onSelected` reconstruction from `_buddyNameFilter` is correct but non-obvious**: `RawAutocomplete` overwrites the controller with the bare selection *before* calling `onSelected`, which is why the code reads the state field instead of the controller. Worth a short comment so it doesn't get "simplified" into a bug later.
- **AND semantics for comma-separated names** ("Alice, Bob" = dives where *both* were present) is consistently implemented and documented across all three paths, which is great. Just flagging it as a product call — the autocomplete encourages adding multiple names, and some users may expect OR. Fine to keep as-is if intentional.
- **Import ordering**: `buddy_providers.dart` was inserted after `tag_providers.dart`, breaking the alphabetical order of the package-import group in `dive_filter_sheet.dart`.
- `row.readNullable<String>('buddy')` is called three times in a row in `_mapSummaryRows`; a local variable would read better (moot if #2 is dropped).
- **Test gap**: there's no test for the comma AND semantics in any of the three matching paths (repository SQL, `DiveFilterSql`, in-memory `apply`) — the widget tests cover the input UX but not the matching. A `dive_filter_state_test.dart` case like `buddyNameFilter: 'alice, bob'` against dives with one vs. both buddies would pin the core new behavior.

ericgriffin (approved):

PR #802, opened 2026-08-01
fix(dive-detail): make "Open Full Page" work on desktop

etlami:
## Problem

In the dive detail pane's ⋮ menu, **Open Full Page** does nothing on desktop.

## Cause

On desktop, a standalone (non-embedded) `DiveDetailPage` deliberately redirects itself into the master-detail pane (`context.go('/dives?selected=:id')`, guarded by `ResponsiveBreakpoints.isMasterDetail`). "Open Full Page" navigated with `context.go('/dives/:id')`, which built a standalone page that immediately hit that redirect and bounced right back into the pane it was opened from — so nothing appeared to happen.

## Fix

`dive_detail_page.dart`:

- Open it with `context.push('/dives/:id?fullpage=1')` instead of `go` — `push` gives a back button, and `?fullpage=1` marks the intent.
- The desktop→pane redirect now skips when `fullpage=1` is present (evaluated after the `isMasterDetail` check, so it only reads the route state where a router is guaranteed).

## Tests

Added to `dive_detail_nav_page_test.dart`: at master-detail width, a plain `/dives/:id` still redirects into the pane (baseline), while `/dives/:id?fullpage=1` opens the full page without redirecting.

Verified locally against Flutter 3.44.8 / Dart 3.12.2: `flutter test` on the dive-detail nav/section suites passes, `flutter analyze` clean, `dart format` clean.


ericgriffin:
Merged main and simplified the approach in 956786e: main's #775 added a canPop guard to the master-detail redirect, so pushed detail pages no longer redirect. With that in place the fix reduces to the menu action using push instead of go. The ?fullpage=1 query-param mechanism is no longer needed and has been dropped. The nav test now drives the actual menu flow (open the pane's ⋮ menu, tap Open Full Page) and asserts the standalone page appears without bouncing back into the pane. Verified the test fails against the old go() behavior and passes with push.

ericgriffin (approved):

PR #826, opened 2026-08-02
feat(profile): continue the NDL overlay as TTS once in deco

etlami:
## What

On the dive profile chart, the **NDL** overlay used to clamp to a flat line pinned at zero the moment a deco obligation began — an uninformative stretch. Worse, because the deco points were only dropped (not broken), the line **bridged straight across the deco region as a misleading diagonal**.

Now the NDL overlay is drawn **only while there is actual no-deco time left** (NDL > 0). Once it's spent, the line **breaks** and the same overlay **continues as TTS** (time-to-surface) over the deco portion, drawn as its own coloured segment so the metric change reads clearly. The tooltip switches to TTS there too.

So the single NDL overlay reads as **NDL while no-deco diving and TTS once a stop is owed** — which also keeps the graph clean without needing a separate always-on TTS line.

## Details

- `_buildNdlLine`: only emits spots where `ndl > 0`; inserts a null spot at deco boundaries so segments don't bridge; straight segments (a spline would overshoot across the breaks).
- New `_buildNdlDecoTtsLine`: plots TTS over the `ndl < 0` deco portion, in the TTS colour, under the same NDL toggle.
- When the standalone TTS overlay is also enabled, it takes over the deco portion — the NDL overlay's TTS continuation (line + tooltip) is suppressed, so **TTS is never drawn or listed twice**.
- Both tooltip paths updated accordingly.

## Note

This is a chart-behaviour change; I validated it visually on a real deco dive (NDL breaks where no-deco time runs out, TTS takes over, no bridging diagonal, single TTS tooltip row). No unit test is added — the change is in the fl_chart line builders, exercised by the existing profile-chart widget tests, which pass.

Verified locally (Flutter 3.44.8 / Dart 3.12.2): `flutter analyze lib` clean, `dart format` clean, dive-log widget suite green.

### DIFF
--- a/lib/features/dive_log/presentation/widgets/dive_profile_chart.dart
+++ b/lib/features/dive_log/presentation/widgets/dive_profile_chart.dart
@@ -4739,8 +4739,17 @@ class _DiveProfileChartState extends ConsumerState<DiveProfileChart> {
 
     final spots = <FlSpot>[];
     for (final i in _decimatedCurveIndices(ndlData)) {
-      // Clamp NDL to display range to avoid gaps that cause Bezier artifacts.
-      // Negative values (in deco) clamp to 0; values > 60 min clamp to 60 min.
+      // Draw NDL only while there is actually no-deco time left. Once it is
+      // spent (zero, or negative in deco) the line simply ends -- a flat line
+      // pinned at zero through the deco phase carries no information. A null
+      // spot breaks the series so it does not bridge straight across the gap.
+      if (ndlData[i] <= 0) {
+        if (spots.isNotEmpty && spots.last != FlSpot.nullSpot) {
+          spots.add(FlSpot.nullSpot);
+        }
+        continue;
+      }
+      // Clamp values > 60 min to the top of the display range.
       final ndl = ndlData[i].clamp(0, maxNdlSeconds.toInt()).toDouble();
       final normalized = ndl / maxNdlSeconds;
       final yValue = band.mapNormalized(normalized);
@@ -4752,9 +4761,9 @@ class _DiveProfileChartState extends ConsumerState<DiveProfileChart> {
       // surface is already cut short by residual loading, which the first
       // sample reflects and a synthetic maximum would not.
       spots: _withFlatSurfaceLeadIn(spots),
-      isCurved: true,
-      curveSmoothness: 0.2,
-      preventCurveOverShooting: true,
+      // Straight segments: a spline across the null-spot breaks would reach
+      // for the gap and overshoot.
+      isCurved: false,
       color: ndlColor,
       barWidth: 2,
       isStrokeCapRound: true,


readme42:
@etlami Thank you for your contribution!

Would you mind adding a before and after screenshot to visualize the proposed changes? Please include the existing NDL and TTS charts as well. 
This helps reviewers a lot. 
Thanks!

etlami:
Thanks @readme42! Before/after on dive #2851, which has a deco obligation.

NDL overlay only
Before — once NDL reaches zero (deco) the line just pins flat at the bottom of the axis for the whole deco phase and carries no information.

1_vorher_nur-NDL

After — the NDL line stops where no-deco time runs out and continues as TTS (magenta) over the deco portion. So the single overlay reads as NDL while no-deco and TTS once a stop is owed.
2_nachher_nur-NDL

NDL + TTS both enabled
Before:

3_vorher_NDL+TTS


After — the NDL line no longer flatlines in deco, and the tooltip no longer lists TTS twice.

4_nachher_NDL+TTS


(A shallow deco stop later in the same dive, for good measure.)
5_nachher_NDL+TTS_deco-stop

readme42:
Thank you very much @etlami, now it's clear to me what you're trying to achieve here. In general I like the idea of avoiding the long "bottom zero NDL line" — it really has no benefit.

A few points I'd like us to think through before this goes in:

- **A TTS line appears while the TTS toggle is off.** NDL and TTS are separate   legend chips. With NDL on and TTS off, the chart draws magenta, the TTS chip still reads "off", and there's no legend entry explaining the colour. You can
  see it in your screenshot 5: a tooltip row labelled TTS with the TTS toggle disabled.

- **NDL and TTS have independent Computer/Calculated source selectors** (`ndlSource` / `ttsSource`, resolved separately in
  `overlayComputerDecoData`). So the continuation can splice a computer-reported NDL into a calculated TTS in what looks like one continuous line, with the second half governed by a selector the user never touched.

- **The multi-computer comparison overlay wasn't updated** — it still clamps the comparison dive's NDL to zero and keeps it curved. With two computers shown, the primary NDL breaks into TTS while the comparison one flatlines at the
  bottom, so one toggle ends up with two different meanings.

Stepping back, the questions I keep coming back to are:

- Why would I, as a technical diver, enable NDL on the graph at all?
- Why would I, as a recreational diver, enable TTS?

I ask because if the honest answer is "tec divers want TTS, rec divers want NDL", then maybe the better fix isn't to chain the two metrics onto one toggle, but to make each overlay behave sensibly on its own — NDL simply ending when it stops
existing, and TTS being the thing you switch on when you dive deco. That would also make the three points above go away. But I might be missing a use case you had in mind, so I'd like your view first.

Happy to keep iterating on this with you either way.

ericgriffin:
TTS and NDL are objectively two distinct things, each with a different class of diver who cares about them, and each with potentially more than one source or way of being calculated/displayed. For those reasons I feel that they should remain separate and distinct in the UI. Merging them makes too many assumptions around what a diver infers from the values. Merging would also require adding a lot of ui configuration complexity to support the different source/calculation combinations of each metric.

etlami:
Agree — chaining the two metrics onto one toggle was the wrong call, and it produced exactly the three issues you flagged (magenta while TTS is off, spliced source selectors, the untouched comparison overlay). Reworked to the simpler behaviour you both suggested: the NDL overlay now just **ends** where no-deco time runs out (no flat zero line, no bridging diagonal), and TTS is left entirely as its own overlay. Each overlay now behaves sensibly on its own, so all three concerns go away. The diff is down to a single ~14-line change in `_buildNdlLine`. Thanks both!

ericgriffin (approved):

PR #870, opened 2026-08-06
Fix large-backup restore: intermittent strict-close failure, UI freezes, and follow-up hazards

ericgriffin:
## Summary

Restoring a large backup froze the UI for long stretches and intermittently failed with `Restore failed: TimeoutException after 0:00:05`, succeeding only after several retries. This PR fixes both root causes and the remaining large-file hazards on the restore path.

### Root cause 1: intermittent failure (the retry-succeeds signature)

`DatabaseService.close(strict: true)` — the close that runs right before the restore's file swap — trusted drift's *graceful* `close()` with a throwing 5-second timeout. Drift's graceful close awaits `streamQueries.close()`, which hangs for as long as **any `watch()` subscription is paused**, and Riverpod 3 auto-pauses the streams of providers nobody is currently listening to. A mid-session restore almost always has at least one paused stream (any previously visited page), so the close hung, the timeout threw, and the restore failed. Retrying eventually worked because the abandoned close future completed in the background once navigation resumed the paused streams — the next attempt found the connection already closed. Setup-wizard restores (fresh session, no paused streams) were unaffected, which is why the failure looked random.

Strict close now mirrors the proven `closeDatabaseForAppShutdown` pattern: a brief graceful attempt, then a direct executor close on timeout. The existing 60-second worker-exit wait remains the real guarantee that the SQLite file handle is released before the swap. Regression test: `restore succeeds while a watch() subscription is paused` (fails on the old code with the exact reported TimeoutException).

### Root cause 2: UI freeze

The backup crypto — the AES-GCM frame loops encrypting the pre-restore safety backup and decrypting `.sbe` restores, plus the Argon2id KDF (64 MiB, t=3) — ran pure-Dart with synchronous file IO **on the UI isolate**. A multi-hundred-MB backup meant the UI thread did roughly twice the file size of pure-Dart AES. All three `BackupCrypto` entry points now run their heavy bodies in `Isolate.run`, passing only sendable inputs (paths, raw key bytes, the secret); `WrongPassphraseException`, `EnvelopeCorruptException`, and `SyncEncryptionRequired` propagate across the isolate boundary unchanged, which the existing crypto suite exercises end-to-end.

### Follow-up hazards fixed

- **Restore ordering:** `restoreFromBackup` ran the pre-restore safety backup *before* decrypting/validating the source, so a wrong passphrase or corrupt file still minted a backup + history entry. It now decrypts and validates first (parity with `restoreFromFile`), with a regression test.
- **Silent migration stall:** restoring an older-schema backup runs the migration ladder during the post-swap reopen with no feedback. Migration progress now flows through `DatabaseService.restore` → adapter → `BackupService` → `BackupOperationState`, so the restore barrier shows `Upgrading database (step X of Y)...`.
- **Whole-file RAM buffering on cloud transfers:** cloud backup upload/download held the entire artifact in memory (`readAsBytes` + a gratuitous `Uint8List.fromList` second copy on upload; full-buffer download on restore). New path-based `uploadFileFromPath`/`downloadToFile` on `CloudStorageProvider`, with buffering fallbacks in the mixin so all implementations and test fakes keep working, and streaming overrides per provider:
  - **iCloud**: same-volume copy + coordinated move — the payload no longer crosses the method channel as one `Uint8List`
  - **Google Drive**: file stream into `drive.Media`; downloads piped straight to disk (the buffering path held three concurrent full copies at peak)
  - **Dropbox**: upload-session chunks / ranged downloads, one 8 MiB slice resident at a time
  - **S3**: multipart upload with abort-on-failure / ranged downloads
  - **Encrypting decorator**: exempt self-framed backups stream through the inner provider untouched; sealed envelopes are opened after a header sniff, and failures never leave ciphertext or a partial file at the destination

  The transfer loops are ports of the contract-tested `MediaObjectStore` implementations, minus resume state (backups are one-shot artifacts).

## Testing

- New tests: paused-watch-stream restore regression, restore-ordering (no stray safety backup on failure), migration-progress surfacing, encrypting-decorator path-based transfers (seal/pass-through/cleanup), mixin fallback behavior
- `flutter test test/features/backup/ test/features/settings/ test/core/services/ test/core/data/` — 2586 passed
- `flutter analyze` — no issues; `dart format .` — clean

### DIFF
--- a/lib/core/database/database.dart
+++ b/lib/core/database/database.dart
@@ -3246,7 +3246,17 @@ class AppDatabase extends _$AppDatabase {
       'AND bottom_time = runtime',
     ).get();
 
+    var processed = 0;
     for (final candidate in candidates) {
+      // On an imported library nearly every dive is a candidate, and during a
+      // migration the executor is a synchronous main-isolate NativeDatabase:
+      // drift's awaits complete in microtasks, so this loop would run as one
+      // unbroken microtask chain that never reaches the timer/vsync queue —
+      // freezing the migration spinner for the whole step. A real event-loop
+      // yield every few dives lets the UI animate while the backfill runs.
+      if (processed++ % 25 == 24) {
+        await Future<void>.delayed(Duration.zero);
+      }
       final diveId = candidate.read<String>('id');
       final runtimeSeconds = candidate.read<int>('runtime');
 
--- a/lib/core/services/cloud_storage/cloud_storage_provider.dart
+++ b/lib/core/services/cloud_storage/cloud_storage_provider.dart
@@ -1,3 +1,4 @@
+import 'dart:io';
 import 'dart:typed_data';
 
 /// Information about a file stored in cloud storage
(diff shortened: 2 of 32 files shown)


ericgriffin:
Addressed both review findings (ed26083):

**1. iCloud staging cleanup — fixed.** The `File.copy` to the `.uploading` staging path now sits inside the same cleanup guard as the coordinated move, so a failed copy (disk full, permissions) can no longer strand a partial staging file. Added a regression test asserting that a failed copy leaves no `.uploading` file behind and that the move never runs.

**2. `graceful.ignore()` on the abandoned close — declined as a false positive.** `FutureExtensions.ignore()` explicitly suppresses a later error on the ignored future; per its dartdoc, an error completing the future after `ignore()` "is also ignored" and is not treated as uncaught. Verified empirically with a `runZonedGuarded` harness: a future that completes with an error *after* `timeout()` fired and `ignore()` was called produces no unhandled-error callback. `unawaited(f.catchError(...))` would be equivalent, not safer.

PR #952, opened 2026-08-10
feat(equipment): add DPV as an equipment type (closes #806)

### LINKED ISSUE #806: Support Scooter as equipment type
I always track on which dives I use scooter, but it is not possible to create "Scooter" as equipment.

I would like to create equipment with type=scooter

I currently use a custom field for this. It works, but not as elegant as scooter being a supported equipment type.

etlami:
## What

Closes #806 — being able to track equipment of type "Scooter". Implemented using the standard term **DPV** (diver propulsion vehicle) for the type name.

- New `EquipmentType.dpv` (display name **DPV**).
- Represented by a scooter icon (`Icons.electric_scooter`) in the two equipment-type icon pickers.
- `enum_equipmentType_dpv` localization added across all eleven locales (kept as the abbreviation "DPV", like BCD/SMB).

## Notes

- Equipment types are persisted by **enum name**, and reads fall back to `other` for unknown values, so this needs **no migration** and doesn't affect existing data.
- The type pickers build from `EquipmentType.values`, so DPV appears automatically; the two exhaustive icon switches are the only code that needed the new case.

Verified locally (Flutter 3.44.8 / Dart 3.12.2): `flutter analyze lib` and `flutter analyze test` clean, `dart format` clean, `flutter gen-l10n` run, equipment / equipment-picker / arb-parity suites green. (One unrelated pre-existing date-dependent flake in `service_schedule_repository_test` — `2026` vs `2025` — is untouched by this change.)


ericgriffin:
Thanks for this. DPV is the right term and the enum-name persistence plus
`other` fallback reasoning is correct, so nothing here needed a migration.

I pushed four commits to the branch to bring the new type in line with how the
other equipment types are supported. Happy to drop or rework any of them if you
disagree.

**1. Equipment-specific fields (`feat(equipment): give DPVs their own attribute fields`)**

Every type except `transmitter` and `other` has curated fields in
`EquipmentAttributeCatalog`, and `attributesFor()` degrades silently to the two
universal attributes when a type has no entry. That meant a DPV could only
record buoyancy and dry weight, which are the least interesting numbers on a
scooter. Added seven:

| key | kind | notes |
| --- | --- | --- |
| `dpv_style` | choice | tow-behind / ride-on / handheld |
| `burn_time_h` | number | dimensionless, like `scrubber_duration_h` |
| `battery_type` | choice | lithium-ion / NiMH / lead-acid |
| `battery_capacity_wh` | number | the figure on the pack, and the one airlines ask about |
| `motor_type` | choice | brushless / brushed |
| `speed_mps` | number | new `speedMps` dimension |
| `depth_rating_m` | number | reuses the key camera and rebreather already share |

Top speed needed a new `AttributeDimension.speedMps`. It is stored canonically
in m/s, matching wind speed and GPS track speed, and displays as km/h or knots
through the existing `UnitFormatter` conversion. It is the first dimension whose
canonical unit is not also its metric display unit, so the metric-identity test
in `equipment_attribute_units_test` now covers the same-unit dimensions and
asserts speed on its own.

14 new `attrLabel_*` / `attrChoice_*` keys across all 11 locales, plus the
resolver arms in `equipment_attribute_l10n.dart`.

**2. Icon switches (`refactor(equipment): one source of truth for equipment type icons`)**

The type-to-icon mapping lived in four files. The two you patched are exhaustive
switches, which is why they were the two that would not compile. The other two,
`equipment_list_content.dart` and `equipment_detail_page.dart`, end in
`default: Icons.backpack`, so the compiler stayed quiet and a DPV would have
rendered as a scooter in the dive-edit picker and a backpack in the equipment
list and detail pages. Those two had already drifted the same way for
rebreather, SMB, reel, knife, hood, gloves and boots.

Extracted `equipmentTypeIcon()` with no default arm, so the next equipment type
is a build error instead of a silent backpack. Where the copies disagreed, the
complete picker-side glyphs won, which changes the list and detail icons for
BCD, the exposure suits and fins.

**3. Import (`feat(import): recognise scooters when importing from MacDive`)**

`MacDiveValueMapper.equipmentType` matches free text and had no scooter branch,
so an imported DPV still landed as `other` even with the type present. Now
matches "dpv", "scooter" and "propulsion". UDDF is untouched: the spec has no
scooter gear tag to map from.

**4. Submodule pointer (`chore: drop the stray libdivecomputer submodule bump`)**

The branch moved `third_party/libdivecomputer` from its base `6fc6ca1` to
`1a47a01`, and main has since moved the same pointer to `08bf592`, so the two
would have conflicted at merge. Restored to the base value so the merge takes
main's.

**Verification:** `flutter test` is green except for one unrelated settings test
that passes standalone both with and without these commits.
`flutter analyze lib test` reports only pre-existing `unawaited_return_in_try_block`
warnings that come from the branch predating the Flutter 3.47 upgrade on main;
none are in touched files. Rebasing on current main should clear them.

ericgriffin (approved):

Issue #690, opened 2026-07-23
Certification agencies: add ACUC, DAN and FFESSM, and make agencies/levels user-extensible


ericgriffin:
## Request

From r/submersion ([post](https://www.reddit.com/r/submersion/comments/1v2t3lm/request_certification_agencies/)):

> Can we please add ACUC (Association of Canadian Underwater Councils) and DAN (Diver's Alert Network) to the list of certification agencies and/or have the ability to add our own like you can with species?
>
> And if it is possible to also add FFESSM, the French diving federation (Fédération française d'études et de sports sous-marins) (https://ffessm.fr/) it would be much appreciated!

Two asks: three specific agencies, and a user-defined escape hatch modelled on the existing species catalog.

## Decisions taken during design discussion

- **Do both in one change**, not "three agencies now, custom later".
- **Go all the way**: `CertificationLevel` becomes table-backed too, not just `CertificationAgency`. Custom agencies get user-defined levels rather than inheriting a generic ladder.
- **Built-ins are add-only**: seeded ladders cannot be renamed, reordered or removed, but a user may append their own level to any agency (e.g. "Ice Diver" under PADI). Appended levels sort after seeded ones so `primaryCertification` ranking stays stable.

## Current state

`CertificationAgency` (`lib/core/constants/enums.dart:92`) is a closed enum: 11 named agencies plus `other`. Each member carries `displayName`, `primaryColor` and `secondaryColor` (consumed by the e-card renderer and wallet card). `CertificationLevel` (same file, line 143) is a closed enum of 49 values with hardcoded English `displayName`.

`CertificationLevelCatalog` (`lib/core/constants/certification_levels.dart`) maps each agency to an ordered ladder plus a shared cross-agency `specialties` list.

Persistence, all as enum-name text:

| Table | Column | Notes |
| --- | --- | --- |
| `certifications` | `agency` (NOT NULL), `level` (nullable) | `database.dart:1733`, `:1734` |
| `courses` | `agency` (NOT NULL) | `database.dart:2571` |
| `buddy_roles` | `agency` (nullable) | `database.dart:1715` |

Blast radius, excluding generated files:

- `CertificationAgency`: 105 references across 23 files in `lib/`, plus 43 test files.
- `CertificationLevel`: 130 references across 15 files in `lib/`, but **93 of those are inside `certification_levels.dart` itself** — only ~37 elsewhere, plus 15 test files.

## Why the three agencies are not three equal-cost additions

[...]

PR #954, opened 2026-08-10
feat(certifications): free-text agency & level for "Other"

etlami:
## Problem

Entering a certification funnels agency and level through fixed dropdowns. When a diver's agency or level isn't in the list, the only option is **"Other"** — which discards the real name (it's stored as the literal `"other"`). The DB columns are already free `TEXT`, so this was purely a UI limitation.

## What this does

Picking **"Other"** for the agency or the level now reveals a **free-text field** to type the actual name. The custom value persists and flows into the **auto-derived certification title** — e.g. agency "TSA" + level "Full Cave" → title "Full Cave" — so an unlisted agency/certification is captured properly instead of lost.

- New nullable `agency_custom` / `level_custom` columns (**v148** migration + `beforeOpen` backstop). The columns were already free TEXT, so existing data is unaffected and no CHECK/enum constraint changes.
- `Certification` entity + repository (both read mappers, both write companions) carry the two fields.
- `certification_title.dart` gains `effectiveAgencyLabel` / `effectiveLevelLabel`; the derived title, `hasDerivedName`, and subtitle use the custom text when the agency/level is "Other".
- The edit form shows the free-text field **only** when "Other" is selected, and the name-field hint previews the derived title live.
- New `certifications_edit_label_agencyCustom` / `_levelCustom` localizations across all locales.

## Deliberately out of scope

This does **not** reintroduce pre-filling the name from agency+level — the title is still derived (and a name that merely repeats agency+level is suppressed) per the 2026-08 change. The name field stays the override.

## Tests

- `migration_v148_certification_custom_test`: fresh schema has the columns; a stranded pre-v148 DB gains them via the `beforeOpen` backstop; no-op when the table is absent.
- `certification_title_test`: added a group for the custom-agency/level cases (effective labels, derived title, and custom-name detection).

Verified locally (Flutter 3.44.8 / Dart 3.12.2): `flutter analyze lib` and `flutter analyze test` clean, `dart format` clean, `flutter gen-l10n` + drift codegen run, certifications / database-migration / arb-parity suites green.


ericgriffin:
Merged current `main` into this branch to clear the conflict. Pushed to `etlami:feat/certification-custom-agency-level`.

This was a **schema version collision**, not just a text conflict, so it needed a renumber rather than a pick-one-side resolution.

### What happened

This branch claimed `v149` for the certifications free-text agency/level migration. While it was open, `main` used **v149** for the duplicate-tag convergence (#1032) and **v150** for the GPS coordinate format (#1041). The migration therefore had to move to **v151**, the next free rung.

### What changed

| Element | Resolution |
|---|---|
| `currentSchemaVersion` | `151` (neither side's value; higher than both) |
| `migrationVersions` | `main`'s 149 and 150 kept, `151` appended, ladder contiguous |
| `onUpgrade` | `main`'s v149/v150 blocks kept; the certifications block moved to `if (from < 151)` and ordered last so steps run ascending |
| `beforeOpen` backstop | kept as-is; stale `v148 backstop` comment relabeled `v151` |
| `_assertCertificationCustomColumns` docstring | v149 references updated to v151 |
| Migration test | renamed `migration_v149_certification_custom_test.dart` to `migration_v151_...`, assertions moved to `greaterThanOrEqualTo(151)` / `contains(151)` |

The renumber is safe because `_assertCertificationCustomColumns()` is already called from the `beforeOpen` backstop as well as the `onUpgrade` step, so a database stamped at any intermediate version still gains the columns.

One small change to the test beyond the rename: it stamped `user_version = 148`, which at v151 would replay `main`'s v149 tag-dedupe and v150 coordinate blocks against a fixture that only has a `certifications` table. I moved the stamp to **150** so it isolates the single v151 step, matching how the other renumbered migration tests here are written.

The `libdivecomputer` submodule also conflicted and was resolved to `main`'s pointer. This branch pointed at `1a47a01` (2026-07-15), an ancestor of both `main`'s `08bf592` and of this branch's own merge-base pointer, so it was rewinding the submodule rather than advancing it.

### Verification on the merged tree

Flutter 3.47.0 / Dart 3.13.0, which `main` moved to today:

- `dart format --set-exit-if-changed lib/ test/` clean, 0 of 3617 files changed
- `flutter analyze` clean, no issues found
- `flutter test` 17,769 passed / 17 skipped / 0 failed
- All three ladder migrations pass together (v149 tags, v150 coordinates, v151 certifications)
- `dart run build_runner build` re-run on the merged tree produced a zero diff, confirming the generated Drift code picked up both `main`'s new `diver_settings.coordinate_format` column and this branch's two certification columns

No changes to the feature implementation itself.

ericgriffin:
Thanks for this, @etlami. Quality is not the reason for this call: the repository plumbing is complete on both read mappers and both write companions, you correctly left sync alone, and the `beforeOpen` backstop is the right pattern.

I am going with [#690](https://github.com/submersion-app/submersion/issues/690) instead, which covers the same request from the same origin. It makes agencies and levels table-backed, mirroring the completed `BuddyRole` to `dive_roles` refactor. That difference is in kind, not size: a custom agency becomes reusable rather than retyped per certification, gets its own level ladder (so `primaryCertification()` can still rank it), and gets a brand color on the e-card. `agency_custom` and `level_custom` would become dead columns the day #690 lands, plus a migration to fold them in.

#690's issue body has the full design if you want a slice of it. Closing in favor of that one.

Issue #765, opened 2026-07-28
Manual date entry doesn't follow localisation format


schtibal:
**Describe the bug**
On Windows, when clicking the pencil "edit" icon to enter a date manually rather than through the date picker, the date format is fixed as MM/DD/YYYY.

**Expected behavior**
Manual date entry should follow the selected localisation date format.

**Screenshots**
Eg, my chosen date format is "DD/MM/YYYY", but I still need to enter MM/DD/YYYY:
<img width="511" height="171" alt="Image" src="https://github.com/user-attachments/assets/49786639-564d-4517-a92b-27df5b6da696" />

**Platform details:**
- OS: Windows
- Version v1.7.0.117

PR #977, opened 2026-08-11
fix(ui): honor the date format setting in the remaining date surfaces

### LINKED ISSUE #964: Date format not updated across the App
Even though I set date format to DD/MM/YYYY on App settings, if I edit dive date or purchase date of an equipment I'm presented with the manual filling of MM/DD/YYYY instead of DD/MM/YYYY.
See screenshot attached.

Image
Image
Image
Image

ericgriffin:
Closes #964.

## What the reporter saw

A diver with **DD/MM/YYYY** selected still got a month-first date picker when editing a dive date or an equipment purchase date. The calendar in their screenshot starts the week on Sunday, which is the `en_US` giveaway.

**Those two flows are already fixed on `main`** by the #765 wrapper in `lib/shared/widgets/app_date_picker.dart`. That fix is not in `v1.7.2.4977` (the current release), so the reporter is seeing a fixed-but-unshipped bug; it ships with 1.7.3.

This PR closes the gaps that were still live on `main`, so the report does not come back through a different door.

---

## Part 1 — pickers that bypassed the wrapper

The bare Material pickers derive their manual-entry format from the ambient `MaterialLocalizations`, so any call site skipping the wrapper silently reverts to MM/DD/YYYY:

- `dive_list_content.dart` — select-dives-by-date-range used a bare `showDateRangePicker`
- `download_step_widget.dart` — the dive computer download cutoff used a bare `showDatePicker`

Changes:

- New `showAppDateRangePicker`, reusing the wrapper's existing `_pickerLocaleFor` / `_fieldHintFor` so all six `DateFormatPreference` values stay defined in one place.
- `showAppDatePicker`'s `initialDate` is now `required DateTime?` — still required, so callers decide deliberately, but the cutoff picker (which may have no date yet) can pass null.
- **New adoption test** that walks `lib/` and fails on any bare `showDatePicker` / `showDateRangePicker`. The doc comment alone did not stop these two from drifting since #765; this makes it mechanical.

## Part 2 — display surfaces with hardcoded patterns

**List rows.** New `UnitFormatter.formatMonthDayWithYear` honors `isDayFirst` and appends the year (full, or two-digit via `shortYear`) outside the reference year. Used by `DenseDiveListTile` and `DenseTripListTile`; the latter became a `ConsumerWidget` to reach `settingsProvider`.

**Tides.** The chart tooltip, times table, section header and the dive-detail tide card hardcoded `EEE, MMM d` / `MMM d`. These widgets already thread `TimeFormat` down as a parameter, so the date preference now rides the same path. Two new `UnitFormatter` statics (`monthDayPattern`, `weekdayMonthDayPattern`) plus `formatWeekdayMonthDay` keep the ordering rule in one place instead of an `isDayFirst` ternary in each widget. The bare `DateFormat('EEE')` axis tick is unchanged — a weekday alone has no order to fix — and there is a test pinning that.

**PDFs.** Every template called `PdfSharedComponents`' three *static* `yyyy-MM-dd` / `HH:mm` formatters, which is exactly why the hardcoding spread across ~20 sites. Those are replaced by an injected `PdfDateFormatter` carrying both the date and time preference, threaded from `export_providers` through `ExportService` into the five templates, the legacy exporter, the course training log and the planner slate. The constructor arguments are required, so a new call site cannot silently fall back to ISO.

**Others.** Data sources section (creation date, plus entry/exit times that also hardcoded 12-hour), Excel statistics sheet, KML document description, and the certificate share image. `formatDateForExport` now reads `DateFormatPreference.pattern` instead of restating all six patterns in a switch.

## What deliberately stays as-is

- **Export filenames stay ISO** (`PdfExportService._fileNameDate`, the CSV/KML/Excel/UDDF filename stamps, backup filenames) so a folder of exports still sorts chronologically. This is now an explicitly commented decision rather than an accident.
- **UDDF and CSV writers are untouched.** Other software parses them; they must stay ISO.
- **The certification card face and card PNG keep `MM/yy`.** A month and year have no day to reorder — widening them changes *what* the card shows rather than *how it reads*, which is a product decision and not this bug. There is a test asserting the compact form survives a day-first preference, so a future sweep does not "fix" it. The e-card's **spoken Semantics label** does take the full ordered date, since it has no width budget and no visual design to preserve.

## Known remaining surface (not in this PR)

An audit turned up a further ~60 sites using locale-driven skeletons (`DateFormat.yMMMd()`, `.MMMd()`, `.jm()`) across certifications, trips, buddies, media, safety and backup. Those follow the *UI language*, not the diver's `DateFormatPreference`, so a day-first diver on an English UI still sees month-first there. That is a mechanical but wide sweep — many are plain `StatelessWidget`s needing a Consumer conversion — and it is better as its own PR than bolted onto this one. Two smaller ones worth naming: `trip_itinerary_tab.dart:308` hardcodes 24-hour, and `file_review_card.dart:48` / `manifest_preview_pane.dart:93` render raw `toIso8601String()` to the screen.

## Testing

- `flutter analyze`: clean. `dart format`: no changes.
- New/updated: range-picker widget tests, the adoption guard, `formatMonthDayWithYear` / `formatWeekdayMonthDay` / pattern-helper unit tests, 8 tide date-order tests, a five-template PDF preference test, provider-level and per-service PDF tests including explicit "filename stays ISO" assertions, dense dive/trip tile tests, data-sources tests, and certification e-card and renderer tests.
- The adoption guard was verified non-vacuous by planting a bare `showDatePicker` call and confirming it fails with the offending `file:line`.
- Full suite: 16,270 tests. The single failure (`backup_encryption_service_test.dart`) passes in isolation, touches nothing date-related, and is the known load-dependent flake.

One coverage note: the planner slate's header date is asserted through a new `planSlateHeaderLine` seam rather than the rendered PDF, because the slate embeds a subset font whose glyph-indexed text the `pdfVisibleText` helper cannot read back. Asserting on the PDF there would have passed vacuously against an empty string.


ericgriffin:
Filed the remaining locale-skeleton sweep as #983 rather than expanding this PR.

Issue #955, opened 2026-08-10
1.5L tank shows as 2L when editing tanks in dive details


JaspalX:
**Describe the bug**
The decimal is rounded in the 'summary' below the tank heading when editing tanks in the dive details screen. This makes 11.1L look like 11L (which is okay I suppose), but also makes 1.5L looks like 2L (which is not okay)...particularly as I also use 2L cylinders. It's a bit confusing.

**To Reproduce**
1. edit a dive and got to add a tank
2. add a 1.5L tank
3. click done
4. the 1.5L tanks is shown as 2L

**Expected behavior**
preserve the number after the decimal so 1.5 is shown, not 2

**Screenshots**
Shows what the tank looks like (2L) vs how it was added (1.5L)

<img width="680" height="652" alt="Image" src="https://github.com/user-attachments/assets/e91f03e5-f3e8-4e88-9d75-1cb67afaaea9" />
<img width="1504" height="228" alt="Image" src="https://github.com/user-attachments/assets/c181c0a2-bc84-4982-b7f3-88231c2635f7" />

**Platform details:**
- OS: macos Tahoe 26.5.2 (25F84)
- Version 1.7.2 (4977)

**Additional context**

PR #1005, opened 2026-08-11
fix(dive-log): dedupe same-computer sources after a sequential merge

mikecarr:
A same-computer sequential merge (`DiveMergeService.apply`, used for "split pair" combines where a computer logged one dive as two separate downloads) carries over both original dives' `dive_data_sources` rows as provenance. When both originals were logged by the same physical computer, this leaves two rows sharing one `computer_id` on the merged dive.

## Root cause

`getProfilesByDataSource()` builds a `computerId -> sourceId` map to attribute each `dive_profiles` row to its source. Two rows sharing a `computer_id` collide in that map — the non-primary row's id silently overwrites the primary's, so every profile point for that computer attributes to the wrong (stale) row. `getDataSources()` has the same one-row-per-computer assumption and returns both rows unfiltered, so the "Sources" chip row shows a duplicate, empty, selectable ghost chip alongside the real source.

Confirmed against dive #955 in production: two `dive_data_sources` rows sharing one `computer_id` — primary with the dive's real stats (30.1m / 99min / 15°C), non-primary "ghost" with stale pre-merge stats (3.3m / 5min / 23°C) — matching exactly what was reported (one entry shows 5 min bottom time on the graph, the other shows the real dive time).

## Fix

Added `_canonicalDataSourceRows()`, a private helper that collapses same-`computer_id` rows down to one (primary-preferred, else earliest-created, matching the existing query ordering both call sites already use). Wired into:
- `getProfilesByDataSource()` — profile points now attribute to the correct source.
- `getDataSources()` — the ghost chip disappears.
- `hasMultipleDataSources()` (via reuse of `getDataSources()`) — was still counting raw rows, so it kept flagging these dives as multi-computer for the 3D scene switcher even after the chip row was fixed; found during final review and fixed the same way.

Display-only: no schema change, no migration, no change to how `DiveMergeService` writes data. Affected dives self-heal the next time they're opened. `getSourceKeysByDiveId()` (needs every raw row for import duplicate-detection fingerprinting) is deliberately untouched.

**Scope note:** unrelated to #645 (cross-device dive-computer BLE identity duplicates, already fixed by #682) — that's a different bug producing duplicate `dive_computers` records, not duplicate `dive_data_sources` rows on one dive.

## Tests

Added a same-computer collision case to `profiles_by_data_source_test.dart` and a new `canonical_data_sources_test.dart` covering `getDataSources`, each with a companion non-colliding case as a regression guard. Extended `dive_computer_data_repository_test.dart` for `hasMultipleDataSources`. Confirmed red before each fix (leaked/misrouted source key, stale row count) and green after.

## Verification

- `flutter test`: 16,874/16,891 passing (16 pre-existing skips). The one failure, `linux_ffmpeg_engine_smoke_test.dart`, is an environment-dependent smoke test in the unrelated media-transcoder subsystem, not touched by this branch.
- `flutter analyze` clean on all changed files.
- `dart format` clean.
- Manually verified against production data: traced dive #955's actual `dive_data_sources`/`dive_profiles` rows through the fixed logic, and confirmed live in a running build — the duplicate source chip is gone and the correct 642-point profile now displays under the one remaining source.

### DIFF
--- a/docs/superpowers/plans/2026-08-11-dedupe-same-computer-dive-data-sources.md
+++ b/docs/superpowers/plans/2026-08-11-dedupe-same-computer-dive-data-sources.md
@@ -1,0 +1,454 @@
+# Dedupe same-computer dive_data_sources rows Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Stop a dive that was created by merging two same-computer segments (e.g. a "split pair" combine) from showing a duplicate, empty-profile source chip alongside the real one with stale stats.
+
+**Architecture:** Add one small private helper to `DiveRepository` that collapses `dive_data_sources` rows sharing a `computerId` down to the canonical one (primary-preferred), and call it from the two read paths that currently assume one row per computer without enforcing it: `getDataSources()` (feeds the "Sources" chip row) and `getProfilesByDataSource()` (feeds the profile chart). Display-only — no schema change, no data migration, no change to how merges write data.
+
+**Tech Stack:** Dart, Drift (SQLite ORM), `flutter_test` with an in-memory `AppDatabase`.
+
+Spec: `docs/superpowers/specs/2026-08-11-dedupe-same-computer-dive-data-sources-design.md`
+
+---
+
+### Task 1: Add `_canonicalDataSourceRows` and fix `getProfilesByDataSource`
+
+**Files:**
+- Modify: `lib/features/dive_log/data/repositories/dive_repository_impl.dart:706-717` (the `sourceRows` query inside `getProfilesByDataSource`)
+- Test: `test/features/dive_log/data/repositories/profiles_by_data_source_test.dart`
+
+- [ ] **Step 1: Write the failing test**
+
+Add this test inside the existing `void main() { ... }` block in
+`test/features/dive_log/data/repositories/profiles_by_data_source_test.dart`,
+after the `'metadata-only sources keep an entry with no points'` test (around
+line 223). It reuses the `dive-1` / `src-a` (primary, computer `dc-a`) /
+`src-b` (non-primary, computer `dc-b`) fixtures already created in `setUp`.
+
+```dart
+  test(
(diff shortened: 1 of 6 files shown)


mikecarr:
Here is an example dive 1) 
image

Dive 2)
image

ericgriffin:
Is there something that we can do to prevent this from happening in the first place?

mikecarr:
Good question. This PR is deliberately a read-time fix (display only, no schema/migration change) — the actual duplicate row creation happens in `DiveMergeService.apply`, step 10 (`dive_merge_service.dart:390-408`), which carries over *both* original dives' `dive_data_sources` rows as provenance even when both originals were logged by the same physical computer. That's the scenario `split_pair` merges hit directly.

A write-time prevention fix would mean collapsing same-computer provenance rows at merge time instead of relying on the read path to dedupe them — different code path, different risk (it's mutating what a merge writes, not just what a query returns), so I kept it out of this PR on purpose rather than widen the diff.

Happy to open a follow-up issue/PR for that if you'd like it tracked separately.

mikecarr:
Addressed the Copilot review: `hasMultipleDataSources()` now counts `DISTINCT COALESCE(computer_id, id)` directly in SQL instead of routing through `getDataSources()`, so it no longer pays for the friendly-name join or the domain-object mapping just to check a count. Same dedup semantics, verified against the existing same-computer collision test plus the full repository/integration/quality-repair suite (371 passing).

ericgriffin:
> Good question. This PR is deliberately a read-time fix (display only, no schema/migration change) — the actual duplicate row creation happens in `DiveMergeService.apply`, step 10 (`dive_merge_service.dart:390-408`), which carries over _both_ original dives' `dive_data_sources` rows as provenance even when both originals were logged by the same physical computer. That's the scenario `split_pair` merges hit directly.
> 
> A write-time prevention fix would mean collapsing same-computer provenance rows at merge time instead of relying on the read path to dedupe them — different code path, different risk (it's mutating what a merge writes, not just what a query returns), so I kept it out of this PR on purpose rather than widen the diff.
> 
> Happy to open a follow-up issue/PR for that if you'd like it tracked separately.

Ok, I'm good with this PR as it's important for clean-up, but if you don't mind creating a follow-up issue, it'd be good to try to prevent this case from occurring in the first place. Thanks.

ericgriffin (approved):

Issue #810, opened 2026-08-01
Dives / Dive / Dive profile – CCR sensor readings are missing.


alpheios-one:
Android v1.7.1.118

The Sherwater dive computer (and others) also analyses the oxygen sensors. I can’t see this reading in the dive profile graph.
The Sherwater Petrel 3, and indeed all models that are CCR-compatible, can analyse and display the voltage of the individual sensors.

This can be helpful when assessing the sensors.
Here is an example from the Sherwater Cloud app:
 
<img width="1620" height="741" alt="Image" src="https://github.com/user-attachments/assets/e97d3627-87d5-4b4b-b3ca-b29c7b6bb672" />

A sample dive log exported from the Sherwater Cloud app in UDDF format:
[Petrel 3 2026-8-1 9-8-48.uddf.txt](https://github.com/user-attachments/files/30620319/Petrel.3.2026-8-1.9-8-48.uddf.txt)

readme42:
@alpheios-one thanks for the report, this turned out to be an interesting one to dig into.

## UDDF and CCR O2 sensor data

UDDF 3.2.3 (published Nov 2018, still the current release) does support individual cells: `<measuredpo2>` may repeat inside a `<waypoint>`, once per sensor, each carrying a mandatory `ref` attribute pointing at a declared `<o2sensor>`.

The unpublished 3.3.0 draft deprecates `<measuredpo2>` in favour of a new `<ppo2>` element using the same per-cell `ref` mechanism — a rename rather than a new capability. It was drafted with AP Diving around 2021 but never published.

## Your dive

Your file contains no `<measuredpo2>` and no `<ppo2>`. The only per-sample oxygen data is `<calculatedpo2>` (262 samples) — the aggregated loop ppO2 the computer votes/averages from the cells, not the cells themselves. There are also no `<o2sensor>` declarations anywhere in the file, so nothing identifies individual sensors.

In short: Shearwater Cloud's UDDF export does not include per-cell readings at all, so the data never reaches Submersion.

I assume you imported this dive from that UDDF file? Could you try downloading the dive directly from the Petrel 3 over Bluetooth instead — per-cell ppO2 traces are already supported and plotted when the data is present, and the direct download should include them.

One note on terminology: your screenshot shows cell voltage in mV. Submersion charts per-cell ppO2, which is what the computer reports over the download interface; the raw mV values aren't exposed there.

## Issues found in our UDDF importer

Investigating this turned up four real bugs on our side, all in the same block:

[...]

Issue #386, opened 2026-06-23
L/min SAC is unreachable for imported dives (silently falls back to bar/min when tank volume is missing)


ericgriffin:
## Summary

A user asked for SAC in litres/minute "and not only bar/minute." Volumetric L/min SAC is **already implemented** (`SacUnit.litersPerMin`, the default `sac_unit` since the 2025-12-20 "SAC calculation improvements"), but it is effectively unreachable for imported dives, so users don't realize it exists.

## Root cause

Volumetric SAC requires the cylinder's water capacity (`L/min = bar/min x tankVolume`). The dive detail view only renders L/min when a tank volume is present:

`lib/features/dive_log/presentation/pages/dive_detail_page.dart` (~line 1604):

```dart
final showLitersPerMin =
    sacUnit == SacUnit.litersPerMin && tankVolume != null;
```

Imported dives (dive computer / Subsurface / UDDF) don't carry cylinder volume, so SAC silently falls back to bar/min. A diver with a large imported library (the reporter has ~160 dives) never sees L/min even though it's the selected/default unit.

## Possible improvements

- Make tank volume easy to populate across imported dives: a default tank-volume setting, a bulk-apply action, and/or auto-populate from tank presets (see #163).
- When L/min is selected but tank volume is missing, show a hint ("enter tank/cylinder volume to display L/min") rather than silently displaying bar/min.

## Related

- #379 - SAC/RMV calculations result in values higher than Subsurface (accuracy)
- #163 - Auto-populate tank presets from equipment inventory

## Source

Reported on Reddit: [Suggestions or not??? (r/submersion)](https://www.reddit.com/r/submersion/comments/1ua82k6/suggestions_or_not/)

> Would it be possible to add the calculation for SAC by liter per minute and not only bar / minute? I use that info a lot in subsurface.

Issue #382, opened 2026-06-22
For dives imported with surface GPS coordinates that don't match to an existing site, offer a way to create a new site


ericgriffin:
For dives imported with surface GPS coordinates that don't match to an existing site, offer a way to create a new site

PR #1034, opened 2026-08-13
feat(libdc): report raw O2 cell millivolts for Shearwater CCR dives

readme42:
Refs #810. Groundwork only: this makes the O2 cell readings available at the libdivecomputer boundary. Nothing is visible in the app yet - the wrapper, schema and chart work follows separately.

## The problem

Shearwater computers log each O2 cell as a raw millivolt reading, plus a per-cell calibration value in the dive header. libdivecomputer multiplies the two to get a partial pressure. When every calibrated cell still holds the factory default calibration value (2100), that product is meaningless, so upstream cleared the calibrated mask and no per-cell sample was reported at all.

A Petrel 3 writes 2100 for all three cells in every log, regardless of whether the diver calibrated. Four logs from the reporter's unit in #810 - spanning a cell 3 replacement on 24 July and a fresh calibration before every dive - carry byte-identical calibration records:

| Dive | Date | Cells flagged calibrated | Calibration values |
| --- | --- | --- | --- |
| #372 | 27 Jun | all three | 2100 / 2100 / 2100 |
| #381 | 21 Jul | all three | 2100 / 2100 / 2100 |
| #382 | 25 Jul | all three | 2100 / 2100 / 2100 |
| #386 | 1 Aug | all three | 2100 / 2100 / 2100 |

It is not the factor the computer used either. Regressing the cells against the ppO2 the Petrel itself recorded gives 0.01992 to 0.02046 bar/mV depending on the dive, so applying 2100 would put every cell 2.6% to 5.4% high, by a different amount each dive. The real calibration is not present anywhere in the log - an exhaustive scan of every record type across all four files finds no other calibration-shaped field, and Shearwater's own database schema (`sensorNCalibrated` / `sensorNCalibrationValue` / `sensorNADCOffset`) confirms libdivecomputer reads the correct field. The firmware simply fills it with a constant.

So the heuristic's conclusion is wrong here, but its intent is right: without a trustworthy anchor there is no honest way to produce a per-cell ppO2.

## The change

The millivolts are measured data and need no calibration to be meaningful - cell health is judged on the raw output, and it is what Shearwater Cloud itself plots. They are now reported whenever the device logs them, while the derived ppO2 is withheld as `NAN` when the calibration cannot be trusted. That preserves the upstream intent rather than overriding it.

Submodule bumped to `08bf592` on `submersion-patches`, which is submersion-app/libdivecomputer#2 merged plus the upstream merge. Mirrored in `patches/0007`, verified byte-identical to the merged commits.

- new `dc_sample_value_t::ppo2.millivolt`, additive so no existing consumer is affected
- the default-calibration heuristic sets a `calibration_trusted` flag instead of clearing the calibrated mask
- the three unrolled emission blocks become a loop setting `millivolt` always and `value` only when trusted
- `millivolt` is assigned at every ppO2 emission site in the other parsers as well

Devices that do store a genuine calibration are unaffected and still report ppO2 - a Petrel 2 at logversion 14 writes a real value (~1932) and never trips the heuristic. The aggregated loop ppO2 reported with `DC_SENSOR_NONE` is unchanged, as is behaviour for every non-Shearwater parser.

## Why nothing changes in the app yet

`libdc_download.c` resets `o2_sensor[]` to `NAN` for every sample, so cells now arriving with `NAN` are behaviourally identical to their not arriving. Verified that all five platform paths guard every one of the six sensor slots against NaN (darwin, both android paths, windows, linux), so nothing reaches Dart where `_cellAverage` would propagate it. A Petrel 3 CCR dive imports exactly as it does today.

`libdc_sample_t` has no millivolt slot and neither does the Pigeon `ProfileSample`, so the field stops at the C boundary until the follow-up.

## Tests

`test_shearwater_o2_millivolt` drives the public parser API over the #810 dive reduced to its raw PNF log blob, with the computer serial anonymised in all three occurrences:

- the aggregated ppO2 is still reported for all 419 samples
- all three cells are reported, 419 samples each
- every per-cell sample carries a millivolt reading, spanning 38..81 mV, with at least one sample showing the three cells differing from one another
- the ppO2 conversion stays withheld

Full native suite 9/9 against the merged submodule. Every assertion was checked by falsification against the finished parser: zeroing `millivolt` fails the range check, ignoring the trust flag fails the withholding check, and copying cell 1's reading to all three fails the distinctness check.

The distinctness and non-vacuity assertions were added after review: as first written the test held vacuously - with no cells reported, `mv_in_range == cells` was `0 == 0` and the untouched `mv_min`/`mv_max` seeds satisfied the range check, so a fully reverted parser still printed PASS.

## Union aliasing

`ppo2.millivolt` lands at offset 16, the same offset as `deco.tts`, and parsers reuse one `sample` variable across their callbacks. Every parser declares that variable inside its per-record loop, so the field is already zero each iteration and no live case was found where an aliasing member is written before a ppO2 emission in the same iteration. The explicit assignments are belt-and-braces for future parsers rather than a fix for a current bug.

## Known, for the follow-up

`millivolt == 0` is in-band: a dead or disconnected cell genuinely reads 0 mV, which is indistinguishable from a device that reports no voltages. That wants deciding before a chart depends on it. Divesoft Freedom also logs cell voltages (`MEASURE_ID_OXYGEN_MV`, hundredths of a millivolt) and could populate the field once there is a log to verify against.


ericgriffin:
I bumped the submodule pointer to the post-merge submersion-patches head so the pin includes the upstream merge. Back over to you...thanks.

ericgriffin (approved):

PR #1110, opened 2026-08-16
fix(statistics): keep Advanced Search on the tab that opened it (#1079)

### LINKED ISSUE #1079: When selecting Advanced Filter from the Statistics page, the view changes context
**Describe the bug**
When the advanced filter is used on the Statistics page, the context is changed to a dive list, not a filtered statistics list.

**To Reproduce**
1. Go to Statistics page
2. Click on the Filter icon
3. Select "Advanced Filter"
4. Select "Apply Filters"

**Expected behavior**
The statistics panel changes to only show statistics for dives matching the filter criteria. Instead, upon closing the filter screen, the context is changed back to the dive list with the selected filter applied.

**Platform details:**
- OS: Windows
- Version: 1.7.3.6027

ericgriffin:
Fixes #1079.

## The problem

The Statistics tab and the dive list keep **separate** filter state (`statisticsFilterProvider` vs `diveFilterProvider`), so the home screen's totals stay unfiltered. `DiveFilterSheet` was parameterized by a target provider and respected that split, but the Advanced Search page it links to was not:

- it read `diveFilterProvider` unconditionally in `initState`
- it wrote back to `diveFilterProvider` on apply
- it then ran `context.go('/dives')`

So from Statistics, the advanced form loaded the wrong filter, saved the user's criteria to the wrong place, and switched the user to the dive list. Two failures in one: wrong state target and wrong navigation target.

## The fix

| File | Change |
| --- | --- |
| `dive_search_page.dart` | Takes an optional `filterProvider`; reads, writes, and labels its primary button from it. Applying a non-dive-list filter pops back to the section that opened the page. |
| `dive_filter_sheet.dart` | Passes its own `filterProvider` as the go_router `extra` on the push. |
| `app_router.dart` | `/dives/search` resolves that `extra` back into the page, falling back to the dive list's filter when it is absent. |

Passing the provider through `extra` (rather than a `?target=` query parameter) keeps the dependency arrow pointing one way: `statistics/` already imports the dive log's filter sheet, and this avoids `dive_log/` having to know statistics providers by name. A missing or malformed `extra` degrades to the previous dive-list behavior rather than throwing, so deep links and the global keyboard shortcut are unaffected.

From Statistics the primary button now reads "Apply Filters" with a filter icon, since nothing is being searched; the dive-list flow keeps "Search" and its existing `go('/dives')` behavior. All three Statistics entry points (mobile app bar, desktop app bar, compact bar) share the one sheet, so the single `extra:` change covers every one of them.

Also replaces the sheet's hardcoded `'Advanced Search'` literal with the existing localized `diveLog_search_appBar` string, so no new ARB keys or translations are needed.

## Tests

New `test/features/dive_log/presentation/pages/dive_search_page_filter_target_test.dart`, 3 cases:

1. applying from Statistics writes the statistics filter, leaves the dive filter untouched, and returns to Statistics
2. applying from the dive list writes the dive filter, leaves the statistics filter untouched, and lands on the dive list
3. the filter sheet forwards its target provider through the push

The two seeds differ (`minRating` 2 vs 4) and the test sets a third value (5) before applying, so each assertion pins down which provider was read and which was written. I confirmed the tests are not vacuous by stubbing the resolver back to the old always-dive-list behavior: case 1 fails, then passes again once restored.

Full `flutter test` exits 0, `flutter analyze` reports no issues, `dart format` is clean. Pushed with `--no-verify` because the pre-push hook resolves `core.hooksPath` to the main checkout and this branch adds a new test file, which trips the hook's known silent-abort path; CI re-runs analyze and test against the pushed branch.

## Out of scope, worth a follow-up

`_applyAndSearch` builds a fresh `DiveFilterState`, so it silently drops the fields the advanced form does not expose: `computerId`, `equipmentAttrKey/Choice/Min/Max`, `buddyId`, and `diveIds`. Set a dive-computer or suit-thickness filter in the sheet, open Advanced Search, apply, and those are wiped. That behavior is pre-existing and affects the dive list identically, and fixing it needs its own decision about what "Clear All" should do with unexposed fields, so it is left out of this PR.

### DIFF
--- a/lib/core/router/app_router.dart
+++ b/lib/core/router/app_router.dart
@@ -327,7 +327,16 @@ final appRouterProvider = Provider<GoRouter>((ref) {
               GoRoute(
                 path: 'search',
                 name: 'diveSearch',
-                builder: (context, state) => const DiveSearchPage(),
+                // Sections with their own filter (Statistics) push this page
+                // with their filter provider as `extra` so the form edits and
+                // applies to that filter (#1079). Every other entry point,
+                // such as a deep link or the keyboard shortcut, gets the dive
+                // list's filter.
+                builder: (context, state) => DiveSearchPage(
+                  filterProvider: state.extra is StateProvider<DiveFilterState>
+                      ? state.extra as StateProvider<DiveFilterState>
+                      : null,
+                ),
               ),
               GoRoute(
                 path: 'match-sites',
--- a/lib/features/dive_log/presentation/pages/dive_search_page.dart
+++ b/lib/features/dive_log/presentation/pages/dive_search_page.dart
@@ -18,10 +18,22 @@ import 'package:submersion/shared/widgets/app_date_picker.dart';
 /// Advanced search page with all filter options in collapsible sections.
 ///
 /// This page provides a comprehensive form for searching dives with
-/// all available filter criteria. When the user taps "Search", the
-/// filters are applied and they're navigated to the dive list.
+/// all available filter criteria. It edits whichever filter the surface that
+/// opened it owns: the dive list's [diveFilterProvider] by default, or the
+/// Statistics tab's own filter when that section pushes the page with its
+/// provider as the route `extra` (#1079). Applying the dive-list filter takes
+/// the user to the dive list; applying any other section's filter returns to
+/// that section, which is already showing the filtered results.
 class DiveSearchPage extends ConsumerStatefulWidget {
-  const DiveSearchPage({super.key});
+  /// Filter this page reads its initial values from and writes back to.
+  ///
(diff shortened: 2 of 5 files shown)


ericgriffin:
Addressed the Codecov report in e587933. The four uncovered lines were exactly the seams the widget tests routed around: they build their own `GoRouter`, so the real `/dives/search` builder never ran, and nothing reached the page's fallback for a section-targeted filter with nothing to pop.

Added three router tests that invoke the `diveSearch` builder directly (pushed provider, no extra, extra of the wrong type) and one page test for the no-pop fallback. Confirmed against `coverage/lcov.info` that every previously missed line is now executed.

Copilot's review generated no comments, and all other checks were green on the first commit.

Issue #1091, opened 2026-08-16
Price of equipment is not saved in the equipment card


urbamax:
Price of equipment is not saved in the equipment card. Devise is saved, but the text box appears empty after the save

- OS: macOS 26.5.2 (25F84)
- Version 1.7.4 (6062)

PR #1112, opened 2026-08-16
fix(seascape): contour level editing, sheet spacing, and blend default

### LINKED ISSUE #1094: 3D Depth Model Dive Site - False Depth Lines
Version 1.7.5.6211

The deep lines are not adopted. 
Or you can't scroll the menu with many entries

Also, I'm not sure if it's meters or feet
It should be the configured unit from the settings. 
For me it would be meters

Image

ericgriffin:
## Summary

Fixes #1094. The custom contour-level editor in the terrain appearance sheet
had three separate defects, all reported together against v1.7.5.6211.

**Typed levels were never adopted.** The depth box committed through
`onFieldSubmitted` only. A decimal keypad has no submit key, so tapping another
row, dismissing the keypad, or closing the sheet discarded the number the diver
had just typed. The reporter's screenshot shows exactly that state: a focused
field holding a value that no code path would ever persist.

**Rows showed stale numbers.** `TextFormField.initialValue` binds its controller
exactly once, and the rows are keyed by list index. Deleting a row therefore
shifted the data up while each field kept its old text, and switching between
metres and feet left the previous unit's numbers on screen.

**The sheet could not be scrolled clear of the keypad.** No
`MediaQuery.viewInsets` padding was applied, so with the keypad up the sheet
kept its full height and the lower rows and the add button had no scroll extent
left to bring them up. That is the "you can't scroll the menu with many
entries" half of the report.

**Units were not shown.** The stored value was already converted correctly
against the diver's depth unit; only the affordance was missing. The box showed
a bare number while the ramp slider directly above it showed `105 m`.

Two follow-ups from reviewing the sheet on desktop are folded in here rather
than split out, since they touch the same widget. **Both go beyond issue
#1094, so they should be judged on their own merits, not waved through as part
of a bugfix:**

- **Spacing.** The custom level rows sat flush, zero pixels between boxes that
  each carry their own border, so the editor read as one dense block. The
  sheet's gaps are now a named scale instead of ad-hoc spacers.
- **Terrain surface now defaults to Blend** rather than Depth colors. This is a
  product default change. `surfaceMode` is unreleased (added after
  `v1.7.4.6062`), so stable installs have nothing stored for it and pick up the
  new default on upgrade; `encode()` always writes the field, so anyone who has
  already chosen keeps their choice. Both halves are pinned by a test. Note
  that means beta users on 1.7.5.62xx keep whatever they have and must flip it
  by hand. No migration, because nothing can distinguish "never chose" from
  "deliberately chose depth".

## Changes

- Level rows own a controller and focus node, and commit on blur (including
  `onTapOutside`, so tapping anywhere else in the sheet counts), on submit, and
  once more on disposal when an edit is still pending. The disposal write is
  deferred through a microtask guarded on the notifier's `mounted` flag,
  because unmounting runs inside a state-locked `finalizeTree`.
- Because a commit can now happen without the diver pressing anything, the
  "is this an edit?" guard compares in DISPLAY space at the precision the box
  shows. The box renders one decimal, so 30 m reads as `98.4` ft and parsing
  that back gives 29.99232 m; comparing in metres would have called that an
  edit and shaved a little off every level on each blur and each dismissal.
- Text that will not parse, or is not positive, restores the stored level
  instead of sitting on screen looking accepted.
- Every mutation reads the current appearance out of the container rather than
  a value captured at build time, so a commit that lands after the row is gone
  cannot resurrect a level the diver just deleted. That read is skipped when
  the notifier is no longer mounted, since teardown can dispose the container
  before the tree unmounts and reading a disposed container throws.
- `didUpdateWidget` re-seeds a row whenever the value it represents actually
  moved. Mid-typing it never moves, so keystrokes are safe.
- `showTerrainAppearanceSheet` pads for the keyboard inset, the same pattern
  the other text-bearing sheets in the app already use.
- Each row carries the diver's own depth unit as a suffix. Because that widened
  the box, the colour dropdown moved into an `Expanded` with `isExpanded: true`;
  the previous trailing `Spacer` could not give space back once a row was full,
  so a long translation of the "default ink" label risked overflowing a phone
  row.

## Test Plan

- [x] `flutter test` passes (17123 tests). The local full run showed four
      failures, all outside this change and all green in isolation:
      `recovery_code_test`, `universal_import_batch_test`,
      `ocr_scan_page_test`, `media_share_helper_test`. The first is a
      pre-existing flake worth filing separately: the EFF wordlist contains
      one hyphenated entry (`yo-yo`) and the test splits the code on `-`, so
      it fails on roughly 0.62% of runs.
- [x] `flutter analyze` passes (project-wide, no issues)
- [x] Manual testing on: **not performed.** The reporter is on Android in a
      German locale; a device pass there would be worth doing before release.

The existing suite pumped the sheet body straight into a `Scaffold`, so it
never built the modal route. That is why all three defects shipped behind six
passing tests. A second harness now pushes the real route, which is where the
keyboard inset and the dismissal commit live. Added coverage:

- a typed level is adopted on blur, in both metres and feet
- a de-locale diver dismissing an untouched sheet changes nothing (this test
  found the rounding drift described above)
- unparseable text reverts to the stored level
- a pending edit is adopted when the sheet is dismissed
- an edit pending when the rows unmount with no focus change is still adopted
- moving to the next row commits the one being left
- the keyboard done action commits the level
- the tap-outside hook drops focus and commits
- choosing a colour keeps the level depth
- deleting a row re-seeds the rows that shift up
- switching depth unit re-seeds the row text
- each row is labelled with the diver's depth unit
- a row lays out on a narrow phone without overflowing
- rows are separated rather than stacked flush (measured exactly 0 before)
- a stored surface choice survives the change of default
- the sheet pads for the keypad, and a long level list still scrolls fully

The scroll test was checked against the unfixed code to confirm it
discriminates: the last control sat at y=696 behind a keypad line of 480.

## Follow-up

The depth box still reads through `double.tryParse` on comma-normalised text,
which is what this field already did. The canonical helpers for that
(`parseUserDecimal` / `formatDecimalForInput` in
`lib/core/utils/number_input.dart`, added for #1091) are not on `main` yet, so
this PR does not depend on them. Once #1091 lands, this field should adopt both
halves; the de-locale guard test above is written so it keeps holding through
that change.

### DIFF
--- a/lib/features/dive_3d/domain/spatial/seascape_appearance.dart
+++ b/lib/features/dive_3d/domain/spatial/seascape_appearance.dart
@@ -54,7 +54,10 @@ class SeascapeAppearance extends Equatable {
   /// contours) as a translucent overlay. Synced per-diver like the rest.
   final bool mapDepthOverlay;
 
-  /// The 3D terrain surface: depth ramp, map imagery, or a blend.
+  /// The 3D terrain surface: depth ramp, map imagery, or a blend. Defaults to
+  /// blend, which reads as terrain rather than a colour chart while keeping
+  /// the depth ramp legible; the drape loads non-blocking, so a diver who is
+  /// offline still gets the plain ramp until it lands.
   final SeascapeSurfaceMode surfaceMode;
 
   const SeascapeAppearance({
@@ -64,7 +67,7 @@ class SeascapeAppearance extends Equatable {
     this.customLevels = const [],
     this.wallAngleDeg = 22.0,
     this.mapDepthOverlay = false,
-    this.surfaceMode = SeascapeSurfaceMode.depth,
+    this.surfaceMode = SeascapeSurfaceMode.blend,
   });
 
   SeascapeAppearance copyWith({
--- a/lib/features/dive_3d/presentation/widgets/terrain_appearance_sheet.dart
+++ b/lib/features/dive_3d/presentation/widgets/terrain_appearance_sheet.dart
@@ -250,8 +250,14 @@ class _LevelRowState extends State<_LevelRow> {
 
   double get _display => widget.level.depthMeters / widget.unitInMeters;
 
-  SeascapeAppearance get _appearance =>
-      _container.read(settingsProvider).seascapeAppearance;
+  /// Null once the scope is gone. The disposal commit runs during teardown,
+  /// which in tests can land after the container has been torn down, and
+  /// reading a disposed container throws. Disposing a container disposes its
+  /// notifiers too, so the notifier's mounted flag is the public proxy for
+  /// "is this container still readable".
+  SeascapeAppearance? get _appearance => widget.notifier.mounted
+      ? _container.read(settingsProvider).seascapeAppearance
+      : null;
 
   @override
   void initState() {
@@ -309,7 +315,9 @@ class _LevelRowState extends State<_LevelRow> {
     final typed = double.tryParse(_controller.text.trim().replaceAll(',', '.'));
     if (typed == null || !typed.isFinite || typed <= 0) return null;
     final appearance = _appearance;
-    if (widget.index >= appearance.customLevels.length) return null;
+    if (appearance == null || widget.index >= appearance.customLevels.length) {
+      return null;
+    }
     final current = appearance.customLevels[widget.index];
     // Compare in DISPLAY space at the precision the box actually shows. The
     // box renders one decimal, so 30 m reads as "98.4" ft and parsing that
@@ -338,7 +346,9 @@ class _LevelRowState extends State<_LevelRow> {
 
   void _write(SeascapeContourLevel? next) {
     final appearance = _appearance;
-    if (widget.index >= appearance.customLevels.length) return;
+    if (appearance == null || widget.index >= appearance.customLevels.length) {
+      return;
+    }
     widget.notifier.setSeascapeAppearance(
       appearance.copyWith(customLevels: _replace(appearance, next)),
     );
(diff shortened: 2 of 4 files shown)


ericgriffin:
Addressed the Codecov report (8 lines missed on the patch). Chasing those lines was worth more than the number suggests: it turned up two things.

**A real fault.** The disposal commit reads the current settings so it cannot resurrect a level the diver just deleted. That read happens during teardown, and a widget test can dispose the container before the tree unmounts, so it was reading a disposed container and throwing. The failure surfaced in whichever test happened to run next, which is exactly the kind of thing that gets written off as flake. Disposing a container disposes its notifiers too, so the row now uses the notifier's `mounted` flag as the public proxy for whether the container is still readable.

**A test that was proving the wrong thing.** `a pending edit is adopted when the sheet is dismissed` passed, but not via the disposal fallback: popping a route unfocuses before teardown, so the focus handler committed first and the fallback never ran. Added a case that removes the rows with no focus change at all, which is the only situation that fallback exists for.

Also covered the colour dropdown (whose write path this PR changed), the keyboard done action, moving between rows, and the tap-outside hook.

One note on the tap-outside test: `TapRegion`'s pointer routing does not reproduce under the test harness, so rather than keep guessing at framework internals that test drives the wired callback directly and asserts the chain this PR is responsible for, unfocus then commit. Called out in a comment so the next reader knows it is deliberate.

Patch coverage is now complete apart from the pre-existing ramp-max slider callback, which this PR does not touch and I left alone to keep the diff on topic.

PR #1143, opened 2026-08-18
fix(healthkit): stop a non-answer on read access zeroing Apple Watch import (#1128)

### LINKED ISSUE #1128: Apple Watch dive import...doesn't find any dives
**Describe the bug**
Using the mobile app on iOS, importing dives from Apple Watch turns up nothing, no dives.
This is using the built in dive logging in healthkit, not the Oceanic or any other dive logging app for apple watch.

**To Reproduce**
1. Go to Transfer/dive computers
2. Select 'import from apple watch'
3. Grant HealthKit Access (note, it asks for HealthKit Access every time I try step 3)
4. See 'HealthKit Access Granted', click next
5. choose a date range that includes as least one dive , click next
6. no dives found

**Screenshots**

**Platform details:**
- OS: iOS
- Version 1.7.2

**Additional context**

ericgriffin:
Fixes #1128.

## Root cause

Apple deliberately refuses to disclose **read** authorization. `HKHealthStore.authorizationStatus(for:)` is only meaningful for share/write access, so the `health` plugin's iOS helper hard-returns `nil` for reads:

```swift
// health-13.3.2 ios/health/Sources/health/HealthDataOperations.swift
private func hasPermission(type: HKObjectType, access: Int) -> Bool? {
    let status = healthStore.authorizationStatus(for: type)
    switch access {
    case 0: // READ
        return nil
```

`HealthKitService.hasPermissions()` collapsed that `nil` with `?? false`, and `fetchDives()` used it as a hard gate:

```dart
if (!await hasPermissions()) return [];
```

So the Apple Watch import could never return a dive on iOS, no matter what the user granted. The same `nil` also explains the reporter's other symptom (the wizard demanding a permission tap on every visit) and a red "HealthKit access not granted" that never cleared in Settings > Data Sources.

"The platform will not say" is not "the platform said no", but a `bool` cannot carry that distinction.

## Fix

Model the answer as a tri-state and gate reads on it:

```dart
enum HealthPermissionStatus { unsupported, granted, denied, undetermined }

extension HealthPermissionStatusX on HealthPermissionStatus {
  bool get canAttemptRead => this != unsupported && this != denied;
}
```

Only a definitive refusal stops the fetch. A failed probe maps to `undetermined` rather than blocking, so a transient error cannot re-break the import.

Status is probed with `WORKOUT` and `HEART_RATE` alone. The native `hasPermissions` returns `false` and short-circuits for any type missing from `dataTypesDict`, so probing with the iOS 16+ dive types would make every older device look permanently denied: the same mistake in a new costume.

## Three more defects on the same path

Each of these also surfaces to the user as "no dives found", which is why one report was four bugs.

| Defect | Effect |
| --- | --- |
| Profile samples hardcoded `depth: 0.0`, built from heart rate alone | Every imported dive would log **0 m max depth** |
| `getHealthDataFromTypes` batches types and rethrows | One unsupported type (the dive series below iOS 16) discarded **every** other type's samples |
| Date pickers return midnight vs the native `.strictStartDate` predicate | Picking the dive's own day as the **end** date excluded it |

On the first: the code comment claiming the package cannot read depth was stale. `health` 13.x registers `UNDERWATER_DEPTH` (metres) and `WATER_TEMPERATURE` (Celsius) in `initializeIOS16Types()`. Depth now drives the profile timeline, and the far sparser temperature and heart rate series are merged onto each depth sample by nearest reading within 30 s, via binary search so a thousand-sample dive stays linear rather than quadratic. Samples sharing a second collapse to the deepest reading. When no depth series exists the old heart-rate-only profile remains as a fallback so the dive still imports.

## UX

The permissions step now requests authorization outright when the platform will not answer. iOS shows its sheet at most once and answers silently thereafter, so requesting is idempotent and repeat visits land straight on the ready state instead of re-asking. The ready screen says plainly that Apple never confirms read access and points at Health > Sharing > Apps > Submersion; a zero-dive fetch result says the same rather than "Proceeding to review...".

Settings > Data Sources reports the tri-state honestly ("HealthKit access is managed in the Health app" for undetermined) and discloses the two newly read data types, per App Store guideline 2.5.1. New strings are translated across all 11 locales.

## Testing

- `healthkit_service_test.dart`: 29 tests, including a regression that stubs the probe to `null` (exactly what iOS returns) and asserts a dive comes back, plus coverage for the depth series, the merge window, per-type query isolation, and the deny/unsupported paths.
- `healthkit_adapter_test.dart`: 114 tests, retargeted at the tri-state, plus whole-day range coverage.
- Full suite: 18881 passed, 19 skipped. `flutter analyze` clean.

**Not verified on hardware.** Reproducing this needs a real Apple Watch Ultra with a logged dive on iOS 17+ (`.underwaterDiving` as a workout type is iOS 17+; the depth and temperature quantity types are iOS 16+). Worth a device check before release.

Note for existing users: the import now requests two additional read types, so iOS will show its permission sheet once more.

### DIFF
--- a/lib/features/dive_import/data/services/healthkit_service.dart
+++ b/lib/features/dive_import/data/services/healthkit_service.dart
@@ -44,7 +44,7 @@ class HealthKitService implements HealthImportService {
       return await _pluginChannel.invokeMethod<bool>(
         'checkIfHealthDataAvailable',
       );
-    } catch (e) {
+    } catch (_) {
       return null;
     }
   }
@@ -53,7 +53,7 @@ class HealthKitService implements HealthImportService {
   Future<bool> get _healthKitExists async {
     try {
       return await _healthDataAvailableProbe() ?? true;
-    } catch (e) {
+    } catch (_) {
       return true;
     }
   }
@@ -109,7 +109,7 @@ class HealthKitService implements HealthImportService {
       return granted
           ? HealthPermissionStatus.granted
           : HealthPermissionStatus.denied;
-    } catch (e) {
+    } catch (_) {
       // A failed probe is not a denial. Reporting "unsupported" here would
       // block the read, which is the exact mistake that broke this import.
       return HealthPermissionStatus.undetermined;
@@ -137,7 +137,7 @@ class HealthKitService implements HealthImportService {
       );
 
       return authorized;
-    } catch (e) {
+    } catch (_) {
       return false;
     }
   }
@@ -208,7 +208,7 @@ class HealthKitService implements HealthImportService {
         startTime: startTime,
         endTime: endTime,
       );
-    } catch (e) {
+    } catch (_) {
       return [];
     }
   }
--- a/lib/features/dive_import/data/services/healthkit_service.dart
+++ b/lib/features/dive_import/data/services/healthkit_service.dart
@@ -73,6 +73,11 @@ class HealthKitService implements HealthImportService {
   /// [HealthDataType.UNDERWATER_DEPTH] and [HealthDataType.WATER_TEMPERATURE]
   /// are the dive-specific series an Apple Watch Ultra records; without them
   /// an imported dive has no profile and a max depth of zero.
+  ///
+  /// Safe to request wholesale below iOS 16, where the plugin never registers
+  /// those two. Its `requestAuthorization` logs an unknown type and moves on,
+  /// leaving it out of the read set, then authorises the types it did resolve.
+  /// An older device is granted workouts and heart rate as usual.
   static const List<HealthDataType> _readTypes = [
     HealthDataType.WORKOUT,
     HealthDataType.HEART_RATE,
@@ -82,9 +87,10 @@ class HealthKitService implements HealthImportService {
 
   /// Types used to probe permission state.
   ///
-  /// Deliberately excludes the iOS 16+ dive types: the plugin reports an
-  /// unknown type as an outright denial, which would make every pre-iOS-16
-  /// device look permanently denied.
+  /// Deliberately narrower than [_readTypes]. Where `requestAuthorization`
+  /// shrugs an unknown type off, `hasPermissions` short-circuits the whole
+  /// probe to false on one, so probing with the iOS 16+ dive types would make
+  /// every pre-iOS-16 device look permanently denied.
   static const List<HealthDataType> _probeTypes = [
     HealthDataType.WORKOUT,
     HealthDataType.HEART_RATE,
--- a/lib/features/dive_import/data/services/healthkit_service.dart
+++ b/lib/features/dive_import/data/services/healthkit_service.dart
@@ -1,20 +1,62 @@
 import 'dart:io';
 
+import 'package:flutter/services.dart';
 import 'package:health/health.dart';
 import 'package:submersion/features/dive_import/domain/entities/imported_dive.dart';
 import 'package:submersion/features/dive_import/domain/services/health_import_service.dart';
+
+/// Answers "can this device do HealthKit at all", or null when unknowable.
+typedef HealthDataAvailabilityProbe = Future<bool?> Function();
 
 /// HealthKit implementation of HealthImportService for Apple platforms.
 ///
 /// Fetches underwater diving workouts and associated data (depth, temperature,
 /// heart rate) from Apple HealthKit. Only available on iOS.
 class HealthKitService implements HealthImportService {
-  HealthKitService({Health? health, bool? isPlatformSupported})
-    : _health = health ?? Health(),
-      _isPlatformSupported = isPlatformSupported ?? Platform.isIOS;
+  HealthKitService({
+    Health? health,
+    bool? isPlatformSupported,
+    HealthDataAvailabilityProbe? healthDataAvailable,
+  }) : _health = health ?? Health(),
+       _isPlatformSupported = isPlatformSupported ?? Platform.isIOS,
+       _healthDataAvailableProbe = healthDataAvailable ?? _askPlatform;
 
   final Health _health;
   final bool _isPlatformSupported;
+  final HealthDataAvailabilityProbe _healthDataAvailableProbe;
+
+  /// The health plugin's own method channel.
+  static const MethodChannel _pluginChannel = MethodChannel('flutter_health');
+
+  /// Ask the platform whether HealthKit exists on this device.
+  ///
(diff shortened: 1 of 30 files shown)


ericgriffin:
Responding to the suppressed review note on `healthKitWholeDayRange` (end bound at `23:59:59.999`).

The observation that the doc comment overpromises is fair, but the suggested remedy does not close the gap it describes. The `health` plugin puts both range bounds on the method channel as `millisecondsSinceEpoch`:

```dart
// health-13.3.2 lib/src/health_plugin.dart:592
'startTime': startTime.millisecondsSinceEpoch,
```

so microseconds are truncated before HealthKit ever sees the value. Both spellings arrive as the same integer:

```
ours      = 2024-01-20 23:59:59.999     ms=1705813199999
suggested = 2024-01-20 23:59:59.999999  ms=1705813199999
identical over the channel: true
```

`23:59:59.999` is therefore the last instant of the day this API can express. Adopting `.999999` would change no behaviour while making the code claim a precision the transport cannot carry.

Fixed the actual defect instead, in f7c95b9: the doc comment now states the millisecond constraint, and a test asserts the two spellings are equivalent over the channel so the extra precision is not reintroduced as a "fix" later.

There is a genuine residual gap of under one millisecond at the native comparison, since `HKSample` dates carry sub-millisecond precision. Closing it would mean passing the next day's midnight as the end bound, which trades a sub-millisecond exclusion for a sub-millisecond over-inclusion depending on whether HealthKit treats `end` as exclusive. Not worth the churn for a dive that would have to start in the final half-millisecond of a day, but noting it in case a reviewer disagrees.

ericgriffin:
Responding to the second suppressed review note, on `isAvailable()` returning `_isPlatformSupported`.

The core concern is right and worth fixing: `Platform.isIOS` is not a HealthKit capability check. Two of the specifics are off, though.

**iPad is not categorically excluded.** HealthKit came to iPad with iPadOS 17, so the real gap is iPadOS 15 and 16 against our iOS 15 deployment target, not "iPad".

**The suggested probe would not detect it.** Treating a throw as unavailable does not work here, because `dataTypesDict` is populated from `HKSampleType.quantityType(forIdentifier:)!`, and those factories return non-nil regardless of whether the device supports the type. So on a device without HealthKit, `hasPermissions` still answers `nil` (read access is never disclosed) rather than throwing, and the probe would report available.

`HKHealthStore.isHealthDataAvailable()` is the honest answer. The plugin already serves it natively under `checkIfHealthDataAvailable`, but its Dart API never calls it, so dc25dc2 asks the plugin's `flutter_health` channel directly behind an injectable probe.

Keeping with the theme of this PR, an unobtainable answer stays unknown and never blocks: only a definitive `false` marks the device unsupported. A missing channel, a renamed handler, or no binding in a unit test all fall back to "available" rather than silently disabling the import.

That result now flows through `permissionStatus()` and `requestPermissions()` as well, so the user-visible improvement is real: a device without HealthKit lands on the wizard's existing "Not Available" screen instead of a "Grant HealthKit Access" button that could never succeed.

Six tests added around the probe (definitive false, unknown, throwing, and the gating of each of the three entry points). Full affected suites pass, `flutter analyze` clean.

ericgriffin:
Responding to the third suppressed review note, on the `HealthImportService` doc calling `HealthKitService` an iOS/macOS implementation while the class says iOS only.

Straightforwardly correct, and pulling the thread found the same false claim in two more places:

- `healthKitServiceProvider` built a service when `Platform.isMacOS`
- `diveImport_healthkit_notAvailableDescription`: "Apple Watch import is only available on iOS and macOS devices", shown to a macOS user on the very screen telling them the feature is unavailable

The `health` package settles it. Its `pubspec.yaml` registers exactly two platforms:

```yaml
flutter:
  plugin:
    platforms:
      android:
        package: cachet.plugins.health
        pluginClass: HealthPlugin
      ios:
        pluginClass: HealthPlugin
```

There is no macOS path, so a service constructed on a Mac could only ever report itself unsupported, and the string was pointing macOS users at macOS.

b6fe817 makes all three agree on iOS, and the user-facing string now names what the import actually needs, an iPhone with the Health app, across all 11 locales. No behaviour change on iOS: `_isPlatformSupported` already defaulted to `Platform.isIOS`, so macOS was landing on the unavailable path regardless. This removes the dead branch and the misleading copy rather than changing what any device does.

Affected suites pass (2106 tests), `flutter analyze` clean.

ericgriffin:
Responding to the suppressed notes from the last three review rounds.

**`catch (e)` with an unused binding** (4 notes). The stated rationale does not apply here: `unused_catch_clause` is not enabled in `analysis_options.yaml` and `flutter analyze` is clean, so no hints were being silenced. The style point stands on its own though, and it earns its keep in this file: every one of those catches is a deliberate swallow, and a PR about a swallowed non-answer zeroing an import is the wrong place to leave catches that look like they inspect the error. Converted in 4cf7695, five of them rather than four, since `requestPermissions` had the same shape.

**Memoising the capability probe.** Correct, and taken in ad41180. The future is memoised rather than the value so concurrent callers share one in-flight round-trip, and `_resolveHealthKitExists` cannot throw, so the cache can never hold a poisoned future. Note the deliberate asymmetry: an unobtainable answer still resolves to available. Caching a refusal would have made a transient probe failure permanent, which is this PR's own bug in miniature. No measured speedup is claimed, and none is intended: a boolean channel call is cheap and was never on a hot path. The point is asking an unchanging question once.

**Requesting the dive types on iOS 15.** This premise does not hold. The unknown-type branch of the plugin's `requestAuthorization` only logs:

```swift
if dataTypesDict[key] == nil, characteristicsTypesDict[key] == nil {
    print("Warning: Health data type '\(key)' not found in dataTypesDict or characteristicsTypesDict")
}
// no return, no throw, no result(false) -- falls through to:
healthStore.requestAuthorization(toShare: typesToWrite, read: typesToRead) { ... }
```

Unresolvable types are simply never inserted into `typesToRead`; the resolved ones are still authorised and the call succeeds. The Dart side only pre-validates write permissions on ECG and heart-event types, which we never request. So on iOS 15 the two dive types are skipped and workouts and heart rate are granted as usual. No denial, no stranded wizard, and no need to split the request.

Worth stating why the confusion is reasonable: this is the exact opposite of `hasPermissions`, which *does* short-circuit the entire probe to `false` on a single unknown type. That asymmetry is why `_probeTypes` is deliberately narrower than `_readTypes`, and only half of it was written down. 9d3a0b3 records both halves, since this has now come up twice.

Issue #1375, opened 2026-08-28
Nothing reclaims on-device caches or temp files; wire the caps and sweeps that already exist


ericgriffin:
## Summary

Nothing in the app runs on any schedule to reclaim disk. The startup sweep (`MediaOrphanBacklogSweep` in `startup_page.dart`) is a database-row sweep for unlinked media, not a filesystem one. Several caches are uncapped and never evicted, a few temp directories are write-only, and the sync tombstone log grows forever for local-only users. There is also no UI where a user can see aggregate disk usage.

Most of this is connecting things the code already intends to do: the eviction functions and cap constants exist, they are just never called or never applied.

## What was measured

Container for a 40-dive dev install: 133 MB. The database is 28 MB of that; a stale 73 MB Sparkle update `.dmg` in `Library/Caches` was larger than the entire database.

Database fragmentation is explicitly NOT the problem. On that database `freelist_count` is 0 and `VACUUM INTO` recovers 1.21 MB (4.3%). Periodic compaction is not worth an exclusive lock contended by two isolates (UI plus the Workmanager headless one). The real storage cost is in the profile sample schema, tracked separately.

### Map tile cache
- `lib/features/maps/data/services/tile_cache_service.dart`: the FMTC store is created with no size cap (`_store!.manage.create()` takes no max length) and no TTL (`FMTCTileProvider` gets no `cachedValidDuration`). Every browsed tile is written under `BrowseStoreStrategy.readUpdateCreate`.
- `removeOldTiles(Duration)` exists in the same file and has zero callers in `lib/` or `test/`.
- Deleting a region from the offline maps page removes the `cached_regions` row but not the tiles.
- Only cleanup is the user "Clear cache" button on the offline maps page.

### `cached_network_image` disk cache
- `lib/features/media/data/network_cache_config.dart`: `kDiskCacheCapBytes = 500 MB` is declared, has a unit test asserting its own value, and is never applied anywhere. The file's own comment says the byte cap needs a custom `BaseCacheManager` that Phase 3c never introduced. Only the in-memory cap is wired.
- Actual disk governance is flutter_cache_manager's defaults (200 objects, 30-day stale), not anything the app sets.

### Video and PDF thumbnails
- `getApplicationSupportDirectory()/Submersion/video_thumbnails` and `pdf_thumbnails` (`media_resolver_providers.dart`, written by `video_thumbnail_service.dart` and `pdf_thumbnail_service.dart`).
[...]

PR #1389, opened 2026-08-29
feat(storage): show on-device storage usage by category

ericgriffin:
Slice A of the storage reclamation program for #1375: a page that reports how
many bytes the app holds on this device, by category.

Read-only. No file is deleted that is not already deletable through existing
UI, no cap is changed, no sweep is scheduled, and `startup_page.dart` is not
touched.

## Why measurement first

#1375 is an audit listing seven remediations across maps, media, backups, sync
and UI. The design doc
(`docs/superpowers/specs/2026-08-29-storage-reclamation-design.md`) splits it
into five slices; this is the first, so the rest are verifiable rather than
merely assertable.

## What the exploration changed

Three findings reshaped the audit, and are recorded in the spec:

- **There is no startup-maintenance host, and one was deliberately abandoned
  twice** (`2026-07-17-data-quality-assistant-design.md:85`,
  `2026-08-06-media-section-phase5.md:589`). New sweeps in later slices follow
  the `MediaOrphanBacklogSweep` shape rather than reviving a task runner.
- **`deletion_log` already has a 30-day GC floor** (`SyncLiveness.gcFloorMillis`)
  and `TombstoneGcDecision.unbounded()` already exists. What it lacks is
  reachability: GC only fires at the tail of a successful cloud sync, so a
  device that never configured sync never gets there. That reframes the
  per-sample tombstone concern as write amplification rather than storage, and
  moves it to its own issue as a wire-format change.
- **Two categories can never be swept**, in any slice. Documents-root exports
  are user-visible in the iOS Files app and share a directory with the
  database; every backup file is a full copy of the database. Both are measured
  and shown, never reclaimed without a user action.

## Design

`StorageCategory` is a descriptor carrying an id, a group, and a `measure()`
that each category implements in its cheapest available form. Measurement cost
is uneven by an order of magnitude, and the existing code already says which is
which: `MediaCacheStore.totalBytes` is a `SUM` over an index,
`TileCacheService.getTotalCacheSize` is FMTC-native, and the rest need a
recursive walk. A `FutureProvider.family` keyed by category id therefore gives
each row independent loading, so the page never blocks on its slowest
measurement and one failing category cannot blank it.

`measure()` returns `Future<int?>`. Null means structurally unmeasurable and is
not the same as zero. Two such cases exist today: an Android SAF backup
location is a `content://` tree URI with no directory to enumerate, and the
tile store throws `StateError` when it never initialized, which is normal since
startup swallows that failure. Reporting either as "0 bytes" would tell the
user their backups or tiles had vanished.

Fourteen categories: the database and local cache database, the three indexed
media cache pools, `staging/` and `transcode/` separately (their bytes are
invisible to the 3.25 GiB LRU budget because neither directory is indexed), map
tiles, network images, video and PDF thumbnails, backups, temporary files, and
Documents-root exports.

## Notes for review

- The page is a new file rather than an addition to `storage_settings_page.dart`,
  which was already at 754 of the project's 800-line ceiling and has one
  coherent job today: where the database file lives. It ends at 766 with the
  entry tile.
- `_formatBytes` was private in `sync_devices_page.dart`; it moves to
  `core/utils/byte_format.dart` and both callers use it. Bytes are not a
  diver-facing unit, so it deliberately does not consult unit settings.
- Deviation from the spec, stated rather than hidden: the spec proposed
  trailing clear buttons on the map tile and network image rows. They are not
  here. `clearAllCache()` in `offline_map_providers.dart:285` also deletes every
  `cached_regions` row, so it is not merely a cache clear, and putting it behind
  a bare button without the confirmation the offline maps page already has would
  be a safety regression. Both clear actions remain reachable from the pages
  that own them, and the slice stays genuinely read-only.

## Testing

26 new tests: 5 for the formatter, 6 for the measurement helpers, 13 for the
inventory, 7 for the page (loading, resolved, unavailable, per-row error,
partial total, and a 360px overflow guard).

Full suite: 22054 pass, 20 skipped. One failure,
`local_file_resolver_test.dart` "a path on the system volume is never probed at
all", is pre-existing and environmental: it passes 44/44 with `TMPDIR` on the
system volume and fails 43/44 with a RAM-disk `TMPDIR` under `/Volumes`.
Nothing in this branch touches that resolver.

Two findings surfaced during exploration that the audit did not have, folded
into slice B rather than dropped: deleting an offline map region removes the
`cached_regions` row but no tiles, despite a doc comment at
`offline_map_providers.dart:262` claiming otherwise; and the per-region size
shown in that UI is `downloadedTiles * 20 * 1024`, a fabricated estimate rather
than a measurement.

Refs #1375. Does not close it: slices B through E remain, and are designed in
the spec.


ericgriffin:
Closed automatically by a branch rename, not deliberately. Continued in #1391 from `worktree-storage-visibility-cleanup`, with all three review comments here addressed:

- The temporary category walked the whole temp directory and so double counted the network image cache, which already has its own row. Now scoped to `picked/` plus loose root files, with a fixture that contains a sibling subtree so the test can actually fail.
- The total header claimed to be final while errored and unmeasurable categories were contributing nothing. All three non-contributing states now hold the partial label.
- The support directory is memoized. The `MediaCacheStore` is still built per call deliberately, since it captures the `LocalCacheDatabase` and a cached one would go stale after a database location migration.

PR #1429, opened 2026-08-31
fix(planner): let the phone results page scroll as one instead of pinning the header

### LINKED ISSUE #1428: Dive planner: phone Results page has pinned chart/tab bar, crowding out plan details
## Problem

On the dive planner's Mission Control page (`PlanCanvasPage`), phone-width
layouts (< 760px) show a "tab deck": the profile chart, the runtime/NDL/CNS/
warnings readouts, and the Plan/Tanks/Setup/Results segmented tab bar are all
pinned in a fixed header, while only the content below it scrolls.

That header eats roughly a third of a typical phone viewport, so on the
Results tab there's very little room left to read the deco schedule, gas
consumption, contingencies, range table, and warnings — the content most
divers actually need to check before a dive.

## Expected behavior

The whole Results page should scroll as one: the chart, tab bar, and stat
tiles should scroll away with the rest of the content instead of staying
pinned, freeing the full viewport for the plan details underneath.

## Root cause

`_buildPhone` in `lib/features/planner/presentation/pages/plan_canvas_page.dart`
built the phone layout as a `Column` with the chart + `SegmentedButton` tab
bar as fixed children, and only wrapped the per-tab body in
`Expanded(child: _phoneTabBody(tab))`. Each tab body was its own independent
`ListView` (for the Results tab, `PlanResultsPane` → `PlanResultsSheet`),
so only that inner region scrolled, and the header above it never moved.

## Fix

Fixed on branch `fix_page_scroling` (commit 09f720e96): reworked
`_buildPhone` to use a single `CustomScrollView` with the chart, tab bar,
and per-tab content as slivers in the same scroll view, so the whole page
scrolls together. `PlanResultsPane`/`PlanResultsSheet` gained an optional
`shrinkWrap` flag so the Results tab's content lays out to its natural
height under the new outer scroll view, while the desktop/tablet layout
(which still needs its own independently-scrolling results panel) is
unaffected.

Verified on a real Android device profile (Waydroid, Pixel 6a dimensions,
API 33) and with the planner test suite (36 tests passing).

dotanalon:
## Summary

On phone-width layouts (<760px), the dive planner's Mission Control page pinned the profile chart, the runtime/NDL/CNS/warnings readouts, and the Plan/Tanks/Setup/Results tab selector in a fixed Column outside the scrollable region. Only the content beneath that header (segments, tank list, setup accordion, or the results sheet) scrolled in its own nested ListView. On a typical phone viewport that header ate roughly a third of the screen, leaving very little room to read the deco schedule, gas consumption, contingencies, range table, and warnings on the Results tab.

Rework _buildPhone in plan_canvas_page.dart to use a single CustomScrollView: the chart, tab bar, and per-tab content are all slivers in the same scroll view, so swiping scrolls the whole page together and the header is no longer pinned.

The Results tab's content (PlanResultsPane / PlanResultsSheet) previously owned its own independent scrolling ListView, nested inside the old Expanded region, with a controller also shared by the always-visible desktop/tablet results pane. Give both widgets an optional `shrinkWrap` flag so the phone tab deck can lay the results content out to its natural height and let the new outer CustomScrollView own the scroll gesture, while the desktop/tablet layout (which still needs its own independently scrolling panel) keeps the previous behavior unchanged.

Verified on a real Android device profile (Waydroid, Pixel 6a dimensions via `wm size`/`wm density`, API 33) that the chart, tab bar, and stat tiles now scroll away together with the rest of the Results tab content, and that scrolling back to the top restores them correctly. Also verified with `flutter analyze` and the planner test suite (plan_canvas_page_test, plan_results_widgets_test, ccr_ui_test, contingency_ui_test, plan_panes_test — 36 tests passing).

## Changes

Fixes #1428 

## Test Plan

- [X] `flutter test` passes
- [X] `flutter analyze` passes
- [X] Manual testing on: waydroid emulator


ericgriffin (approved):
👍

Issue #1399, opened 2026-08-30
MacDive import: dive profiles not recovered for Suunto computers


trustthegoose:
**Describe the bug**
Importing a MacDive SQLite database containing dives from Suunto computers does not produce dive profiles, even though MacDive's `ZRAWDATA` column is actually populated for at least some Suunto models. The importer's existing analysis ([`docs/import-formats/macdive-zsamples.md:5`](https://github.com/submersion-app/submersion/blob/main/docs/import-formats/macdive-zsamples.md#L5)) assumed `ZRAWDATA` is only ever populated for Shearwater computers, based on a corpus that never included any Suunto dives - so that assumption was untested, not verified, for Suunto.

**To Reproduce**
1. Import a MacDive database containing dives logged on a Suunto dive computer (e.g. EON Steel, EON Steel Black, Cobra).
2. Complete the import.
3. Open one of the imported Suunto dives.
4. Observe no depth/profile chart, despite the dive otherwise importing successfully.

**Expected behavior**
Suunto dives with usable `ZRAWDATA` should have their profile decoded and imported, the same way Shearwater dives already are.

**Platform details**
- OS: macOS
- Version: v1.7.7 (build 124)

**Additional context**
Suunto's `ZRAWDATA` (for EON Steel, EON Steel Black, Cobra at least) is stored uncompressed, already in libdivecomputer's native format (SBEM for EON Steel family, Vyper header for Cobra) - unlike Shearwater's stream, it needs no decompression pass before being handed to libdivecomputer's parser. A fix is in progress; PR to follow.

trustthegoose:
For context: this is a follow-up gap left by #912 (fixed by PR #961), which shipped Shearwater \`ZRAWDATA\` decoding and explicitly scoped it to Shearwater only, noting "Oceanic and manual-entry dives have nothing to decode and still need MacDive's XML export for profiles." That scope limit was never actually tested against Suunto data - no Suunto computer was in the investigation corpus - so it turned out to be wrong for at least three Suunto models. Fix incoming in #1400.

PR #1494, opened 2026-09-03
fix(import): stop double-converting MacDive O2/He gas percentages

### LINKED ISSUE #1493: MacDive import: nitrox/trimix O2 and He percentages are 100x too high
**Describe the bug**
Tanks imported from a MacDive SQLite database show wildly incorrect gas mixes for any non-air fill - e.g. an EAN32 tank imports as 3200% O2 and -3100% N2. This corrupts OTU/CNS calculations for the dive (and any dive-log statistics that depend on gas mix).

`MacDiveDiveMapper` (`lib/features/universal_import/data/services/macdive_dive_mapper.dart`) assumes `ZGAS.ZOXYGEN`/`ZGAS.ZHELIUM` are stored as 0-1 fractions (e.g. 0.32 for EAN32) and multiplies by 100 to reach `GasMix`'s percent scale. That assumption was introduced in ea4349a7 (April 2026) and was never checked against a real gas table - only against a synthetic fixture that encoded the same (wrong) assumption.

**To Reproduce**
1. Import a MacDive database containing at least one tank filled with a non-air gas (nitrox, trimix).
2. Open the imported dive's Gas & Gear section.
3. Observe O2/He percentages far outside a valid range (0-100%), and correspondingly nonsensical N2 (which is derived as `100 - O2 - He`).

**Expected behavior**
Gas mix percentages should match MacDive's own displayed values (e.g. 32% O2 for EAN32), and downstream OTU/CNS calculations should be correct.

**Platform details**
- OS: macOS
- Version: v1.7.7 (build 124)

**Additional context**
Confirmed against a real MacDive database's `ZGAS` table: every row stores a whole percent already, not a fraction - including `"Trimix 21/35"` at `ZOXYGEN=21.0`/`ZHELIUM=35.0`, which matches its own name exactly only if read as percent. Fix incoming.

trustthegoose:
## Summary

Tanks imported from a MacDive SQLite database show gas mixes inflated by 100x for any non-air fill - e.g. an EAN32 tank imports as 3200% O2 and -3100% N2, which corrupts OTU/CNS calculations for the dive.

`ZGAS.ZOXYGEN`/`ZGAS.ZHELIUM` were assumed to be 0-1 fractions (0.32 for EAN32) and scaled by `*100` to reach `GasMix`'s percent scale. That assumption, introduced in ea4349a7 (April 2026) alongside the SQLite gas-mix cast fix, was never checked against a real gas table - only against a synthetic test fixture that encoded the same unverified assumption.

A real MacDive database's `ZGAS` table shows every row already stores whole percent: 21.0 for air, 32.0 for EAN32, and 21.0/35.0 for a gas literally named "Trimix 21/35" - which only makes sense read as percent, not as a 0-1 fraction.

Found while manually validating #1399 / #1400 (the Suunto `ZRAWDATA` profile-decoding fix) against a real MacDive database: once Suunto dive profiles started decoding, their OTU calculations exposed this separate, pre-existing bug. It is unrelated to the Suunto fix itself and affects any MacDive import with a non-air gas, Suunto or otherwise.

Fixes #1493

## Changes

- `_buildDiveMap`'s tank mapping now passes `gas.oxygen`/`gas.helium` straight through to `GasMix` instead of multiplying by 100.
- Default fallback (used when a tank has no linked gas) changes from `0.21` to `21.0` to stay a valid percent.
- Updated the synthetic test fixture (`test/fixtures/macdive_sqlite/build_synthetic_db.dart`) to store `ZOXYGEN` as whole percent (32.0), matching real MacDive data, and corrected the stale test comment.

## Test Plan

- [x] `flutter test` passes
- [x] Manual testing on: macOS - reproduced the bug against a real MacDive database (EAN32 tank showing 3200% O2), applied the fix, and confirmed the same tank now imports as 32% O2

### DIFF
--- a/lib/features/universal_import/data/services/macdive_dive_mapper.dart
+++ b/lib/features/universal_import/data/services/macdive_dive_mapper.dart
@@ -830,9 +830,13 @@ class MacDiveDiveMapper {
           entry['runtime'] = Duration(seconds: t.duration!.round());
         }
         if (t.supplyType != null) entry['supplyType'] = t.supplyType;
+        // MacDive's ZGAS.ZOXYGEN/ZHELIUM store whole percent (32.0 for
+        // EAN32), not a 0-1 fraction - confirmed against a real MacDive
+        // database, where every gas row (including "Trimix 21/35" at
+        // ZOXYGEN=21.0/ZHELIUM=35.0) matches its percent-based name exactly.
         entry['gasMix'] = GasMix(
-          o2: (gas?.oxygen ?? 0.21) * 100.0,
-          he: (gas?.helium ?? 0.0) * 100.0,
+          o2: gas?.oxygen ?? 21.0,
+          he: gas?.helium ?? 0.0,
         );
         tanks.add(entry);
       }
--- a/test/features/universal_import/data/services/macdive_dive_mapper_test.dart
+++ b/test/features/universal_import/data/services/macdive_dive_mapper_test.dart
@@ -300,7 +300,7 @@ void main() {
       expect(tank.containsKey('workingPressureBar'), isFalse);
       // gasMix must be a `GasMix` object, not a Map — UddfEntityImporter does
       // `t['gasMix'] as GasMix?` and a Map cast would throw at runtime.
-      // MacDive stores oxygen as a fraction (0.32); GasMix.o2 is a percent.
+      // MacDive stores oxygen as a whole percent (32.0), matching GasMix.o2.
       expect(tank['gasMix'], isA<GasMix>());
       final gasMix = tank['gasMix'] as GasMix;
       expect(gasMix.o2, closeTo(32.0, 0.01));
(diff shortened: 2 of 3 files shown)


trustthegoose:
Verified in a full local build against a real MacDive database: gas percentages now import correctly (e.g. the previously-broken EAN32 tank now shows 32% O2 instead of 3200%).

trustthegoose:
Follow-up from further testing: this same fix also resolves an inflated RMV/gas-consumption readout on affected dives. Submersion's SAC calculator uses a real-gas (Z-factor) correction that factors in O2/He percentage as a refinement over the simpler ideal-gas formula, so feeding it a badly out-of-range gas percentage threw off that correction too. With the gas mix importing correctly, RMV comes out correct as well - no separate change needed there.

ericgriffin (approved):

Issue #1481, opened 2026-09-02
Ratio IX3M2 PRO dives imported as 0 m / 0 min after firmware 5.2.11/016 update


Cicatr1x:
Describe the bug

After updating a Ratio IX3M2 PRO to firmware "5.2.11/016", dives imported into Submersion are incorrectly recognized as having 0 m depth and 0 minutes duration.

The issue is caused by a change in the sample record type used by the new firmware. The affected samples use record type "0x800E", which is not recognized by the version of "libdivecomputer" currently used by Submersion. As a result, the samples are skipped during parsing.

The issue has already been identified and fixed upstream in "libdivecomputer".

To Reproduce

1. Use a Ratio IX3M2 PRO.
2. Update the dive computer to firmware "5.2.11/016" ("50211016").
3. Record a dive.
4. Download/import the dive into Submersion.
5. The dive is imported with 0 m depth and 0 minutes duration.

For comparison, dives recorded with the previous firmware "5.2.6/016" or "5.2.9/016" ("50206016/50209016") are imported correctly.

Expected behavior

Dives recorded with firmware "5.2.11/016" should be parsed correctly, including their depth profile, duration, and samples.

Screenshots

Not applicable.

Platform details:

- OS: [e.g. iOS, macOS, Windows, Android]
- Submersion version: [e.g. vX.X.X]
- Dive computer: Ratio IX3M2 PRO
- Firmware: "5.2.11/016" ("50211016")

Additional context

The issue has already been fixed upstream in "libdivecomputer":

https://github.com/libdivecomputer/libdivecomputer/commit/8d50dc6c983933887dc84063303fa289415c5495

The upstream commit is titled "Fix the detection of the info record" and fixes the parsing of the changed record types used by the affected firmware.

Could the "libdivecomputer" submodule used by Submersion be updated to a revision containing this fix?

No changes to Submersion itself may be necessary if the fix can be picked up simply by updating the "libdivecomputer" submodule.

Cicatr1x:
Thanks for the Implementation, works Like a charm.

Issue #1419, opened 2026-08-31
SEAC Tablet no BLE discover


Cicatr1x:
**Describe the bug**
My SEAC Tablet is not discoverable in the android app, i tried it paired and unpaired but it wont be found as a dive computer. Syncing via the USB cable is no problem just not via BLE.

**To Reproduce**
1. Go to 'More -> Übertragung'
2. Click on 'Tauchcomputer'
3. Click on 'Neuen Computer verbinden'
4. SEAC Tablet not discovered

**Expected behavior**
SEAC Tablet dive computer discovered and usable for sync,

**Platform details:**
- OS: Android
- Version: v1.7.6.7161

Cicatr1x:
Discovery and pairing works now with the latest Beta Version 1.7.7.7844. So its just the issue with the communication left.

PR #1527, opened 2026-09-03
feat(dive computer): report the Cressi Leonardo deco obligation (relands #342)

ericgriffin:
Relands the substance of #342 by @essengee, which has been open since June and
is now 900+ commits behind main. Closes #342.

## What #342 found, and what is left of it

Two real discoveries about the original Cressi Leonardo, both still valid:

- it logs a **decompression obligation in bit 11** of each 16-bit sample word,
  which nothing ever read, so neither the obligation nor its start and end
  reached the app;
- it grades its **ascent rate on four levels** and only level 3 sounds the
  alarm, so raising `SAMPLE_EVENT_ASCENT` for every non-zero level marked most
  of an ordinary ascent as a warning.

#342 also fixed the off-by-one in the platform event-code tables. Main fixed
that independently on 2 September with one shared `libdc_event_type_name()`
table, so those four binding files are **not touched here**; merging #342's
version would have reverted the consolidation.

## The parser change

Reports the obligation **on each transition** rather than on every sample,
which is what the wrapper's deco carry-forward in `libdc_download.c` expects.
A dive that never incurs an obligation reports nothing at all and stays
*unknown*, rather than asserting `DC_DECO_NDL` with a zero time on every
sample.

That distinction is the reason this is not just #342's patch rebased. The
Leonardo measures no numbers to go with the obligation: no stop depth, no stop
time, no remaining no-stop time. Reporting `NDL = 0` for a whole recreational
dive would be read literally by `parsed_dive_mapper` (`ndl: s.decoType == 0 ?
s.decoTime : null`) and offered to the profile as the computer's own NDL
series.

So the profile overlay gets the rule it already applies to TTS: a series that
is zero from end to end counts as unreported, and the app's calculated curve
stands rather than being overdrawn with zeroes. NDL and ceiling now match TTS.

Bit 11 and the ascent grading are **confirmed on the Leonardo only**. The
Giotto, Newton and Drake share this parser and keep their previous behaviour,
pinned by a test.

## Deco stop start and end

Both ends of a stop arrive as the same `SAMPLE_EVENT_DECOSTOP`, separated only
by `SAMPLE_FLAGS_BEGIN` (1) and `SAMPLE_FLAGS_END` (2). Without reading the
flag the end of a stop persisted as another stop starting. Download and
re-parse now both read it, through one shared constant in
`libdc_sample_units.dart` rather than a copy each.

## Water temperature

The Leonardo reports it in the dive header and never as a sample. The download
path only ever derived it from samples, so it was lost, while re-parse already
preferred the header value. Download now matches re-parse, and a missing
header value still cannot blank a temperature the samples do carry.

## CI: the submodule pointer gate

A submodule pointer is a single sha with no file diff, so a pull request that
moves libdivecomputer backwards reverts every fork commit in between while
showing a reviewer nothing at all to look at. #342 sat with a pointer that
would have silently undone the Seac Tablet BLE name filter (#1419) and the
Ratio iX3M 2 APOS5 record-type fix (#1481). Nothing else in the pipeline could
catch it: the app still builds, and the reverted commits are drivers no CI job
exercises.

The new `submodule-pointer` job asserts the pointer never contains less than
the base branch already contains. It may run ahead of the fork's
`submersion-patches` tip (a patch is pushed to the fork before the app PR that
adopts it merges) or stand still; it may not go backwards. Run against #342 as
it stands, it fails and names the six commits that would be lost.

## libdivecomputer

`submersion-patches` fast-forwarded `ac84a2d..5955815`, so the patch line stays
linear and no branch diverges. The pointer here moves forward to `5955815`, and
the change is mirrored in `patches/0008-cressi-leonardo-deco-and-ascent-fix.patch`.

## Testing

- `packages/libdivecomputer_plugin/test/native/test_cressi_leonardo_deco.c`:
  five cases through `libdc_parse_raw_dive`, the same entry point downloads and
  re-parses use, so the parser change and the wrapper's carry-forward are
  covered together. Verified to fail against the unpatched parser.
- Dart: deco start/end mapping on both the download and re-parse paths, the
  header/sample water-temperature order including the no-clobber case, the
  zero-series rule for NDL and ceiling, and that a start/end pair still
  defaults the dive to technical.
- `flutter analyze --fatal-infos` clean, `dart format` clean, full suite green.

## Note for existing users

Already-downloaded Leonardo dives keep their stored events and samples. The
dive computer detail page's "Re-parse all dives" rebuilds them from the stored
raw data.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1194,6 +1194,82 @@ jobs:
   # `changes` is likewise listed: on a docs/CI-only change it reports code=false
   # and the whole pipeline SKIPS (an accepted green), but if the detector itself
   # FAILS we must not silently skip-then-green, so its failure must reach here.
+  # A submodule pointer is a single sha with no file diff, so a pull request
+  # that moves libdivecomputer backwards reverts every fork commit in between
+  # while showing a reviewer nothing at all to look at. PR #342 sat with a
+  # pointer that would have silently undone two shipped driver fixes. Nothing
+  # else in the pipeline can catch this: the app still builds, and the reverted
+  # commits are drivers no CI job exercises.
+  #
+  # The invariant is one-directional on purpose. The pointer may run AHEAD of
+  # the fork's submersion-patches tip (a patch is pushed to the fork before the
+  # app pull request that adopts it merges), and it may stand still. It may
+  # never contain less than the base branch already contains.
+  submodule-pointer:
+    name: libdivecomputer pointer
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+          submodules: true
+
+      - name: The libdivecomputer pointer must not go backwards
+        env:
+          EVENT: ${{ github.event_name }}
+          PR_BASE: ${{ github.event.pull_request.base.sha }}
+          PR_HEAD: ${{ github.event.pull_request.head.sha }}
+          PUSH_BEFORE: ${{ github.event.before }}
+          PUSH_AFTER: ${{ github.sha }}
+        run: |
+          set -euo pipefail
+          sub=packages/libdivecomputer_plugin/third_party/libdivecomputer
+          if [ "$EVENT" = "pull_request" ]; then
+            base="$PR_BASE"; head="$PR_HEAD"
+          else
+            base="$PUSH_BEFORE"; head="$PUSH_AFTER"
+          fi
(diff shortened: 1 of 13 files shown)


essengee:
Thank you for the work on this @ericgriffin. That original PR of mine was really premature - I should have just left it local for my own personal use until I could get it more stable - and I simply have not had time to get back to it.

Submersion is coming along very, very well. Definitely my app of choice now for logging. Hoping to be able to contribute more in the future...

PR #1566, opened 2026-09-05
feat(navigation): add reorderable Favorites section to the sidebar

dotanalon:
## Summary

Adds a **Favorites** section to the wide-screen navigation rail, directly under **Home**. Users can star any rail destination to move it into Favorites, drag favorites into their preferred order, and un-star a favorite to send it back to its original place in the rail.

Starring **moves** a destination rather than duplicating it: the entry leaves its original section and is rendered exactly once, under Favorites. The order is persisted per device so the sidebar comes up the same way on the next launch.

## Changes

- **Favorites block under Home** (`NavFavoritesSection`): lists starred destinations in the user's order. Extended rail shows a header plus a muted "Star an item to add it here" hint when empty; the collapsed rail shows favorite icons only and renders nothing when the list is empty.
- **Star to move, un-star to restore** (`NavStarButton`): a star on each rail destination adds it to Favorites and removes it from the rail proper; the filled star on a favorite moves it back. If every destination is starred, the trailing divider is dropped so the block does not end in a rule with nothing under it. Home and the phone-only "more" sentinel are pinned and cannot be starred.
- **Drag reordering** via `ReorderableListView`: trailing drag handle when the rail is extended on desktop, long-press drag otherwise.
- **Route highlight follows the item** whether it is rendered in the rail or under Favorites.
- **Persistence** (`NavFavoritesNotifier`): ordered ids stored in SharedPreferences under `SettingsKeys.navFavoriteIds`, seeded synchronously so the first frame is correct. `normalizeNavFavoriteIds` drops unknown, pinned and duplicate ids while preserving order, so a destination removed in a later release simply disappears from Favorites.
- **Shared rail tile** (`NavRailTile`, `NavRailIcon`, `NavRailLabel`, `NavRailLeading`) extracted from `main_scaffold.dart` so Home, the rail destinations and the favorites all render through one widget.
- **Localization**: `nav_favorites`, `nav_favorites_emptyHint` and the star / drag tooltips added to every locale.

## Test Plan

- [x] `flutter test test/shared/widgets/nav test/shared/widgets/main_scaffold_test.dart` passes (77 tests; 17 cover the favorites sidebar)
- [x] Full `flutter test`: 21626 passed, 22 skipped, 3 failed. The 3 failures (`service_schedule_repository_test` since-filter year, `deco_filter_providers_test` eventOnly axis, `linux_ffmpeg_engine_smoke_test` output-size bound) fail identically on the base commit `2f0d5ed2e` with this branch absent, so they are pre-existing and not touched here.
- [x] `flutter analyze` passes (no issues)
- [x] `dart format .` produces no changes
- [ ] Manual testing on: desktop (Linux) rail, extended and collapsed

New tests:
- Notifier: add, remove, toggle, reorder, persistence round-trip, stale and duplicate id filtering.
- Sidebar widget: Favorites renders under Home, empty hint, starring moves an entry (rendered once, under Favorites), un-starring restores it, end divider hidden when all are starred, tapping a favorite navigates, current-route highlight applies to the favorited entry.

## Screenshots

To be added after manual verification on desktop.


ericgriffin:
Thanks for the work on this, but I don't see the point in having a Favorite section when all of the options are already visible in desktop mode. It made it go from "all options are visible", to "now I have to scroll to see everything", and also adds permanent clutter to the window. I see the value in re-ordering (in a settings menu only), but favorites and reordering when everything is already visible is redundant (just order your most used items to the top).
Screenshot 2026-09-05 at 11 26 59 AM

We can revisit this idea later if/when there are more options, but let's table the "Favorites" section for now.

In general for UI changes I think it'd be best to have a discussion around them first, so viability and design details can be worked out before implementation. That'll help us build a more consistent and polished product.

Issue #1568, opened 2026-09-05
update error on windows?


wgreene310:
I installed submersion on windows a week or so ago. Seemed to
be working fine. Now I get this error message: 

<img width="985" height="170" alt="Image" src="https://github.com/user-attachments/assets/2056c4df-b49b-4fe6-9d50-c2bbcd01f74f" />

OK, I download Submersion-v1.7.6.7161-Windows-Setup.exe.Update seems to go OK but still get the same error. I can see the .exe filehas not changed.
Very frustrating.

ericgriffin:
The only way this could happen is if you enabled the beta update channel (or manually installed a beta build) which would upgrade your database to a newer schema, and then went back to the stable release. At this point if you want to remain on the stable release channel you'll need to install a beta build, export a backup, reinstall stable release, and restore. Or just stay on the beta build channel.

Latest beta: https://github.com/submersion-app/beta-builds/releases/tag/v1.7.7.8064

ericgriffin:
@wgreene310 I did some more research into this. What happened is that at some point a beta build of Submersion ran on your machine. Beta builds upgrade the dive log database ahead of the stable release, and the stable build then refuses to open it rather than risk damaging your data. The error tells you to download the latest version, which is genuinely useless advice here, because the latest stable version is the one that just refused to open it. That wording is my bug and I'm fixing it.

Reinstalling won't help either. Your dive log lives outside the install folder, so replacing the .exe never touches it.

The good news is that Submersion automatically saved a full copy of your log right before that upgrade, so you can almost certainly get back to where you were.

Close Submersion completely, then:

1. Open C:\Users\<you>\Documents\Submersion
2. Rename submersion.db to submersion-NEWER.db
   Keep this file. It has any dives you logged recently. If you also see
   submersion.db-wal or submersion.db-shm, rename those the same way.
3. Open the Backups folder inside it. Look for files named like
   20260830-141205123-v175-v186.db
   The two numbers at the end are "upgraded from" and "upgraded to". You want
   the one whose FIRST number is 175 or lower, and if there are several, the
   lowest one.
4. Copy that file into C:\Users\<you>\Documents\Submersion and rename the copy
   to submersion.db
5. Start Submersion.

It should open normally. Anything you logged while the newer build was running will not be in it, but it isn't lost either, it's in the submersion-NEWER.db file you set aside. If you want those dives back, install the latest beta, which will open that file, and export before deciding what to do.

[...]

Issue #1591, opened 2026-09-05
Beta releases are labelled Latest and their assets are indistinguishable from stable


ericgriffin:
## Summary

`submersion-app/beta-builds` is linked publicly from `README.md` and
`docs/guide/installation.md`, and for Linux packages it is currently the only
place to get a `.deb` or `.rpm`. Nothing on that page tells a visitor the builds
are not stable:

| Signal a visitor would rely on | What the page shows today |
| --- | --- |
| Pre-release badge | `isPrerelease: false`, so GitHub labels the newest release **Latest** |
| Asset filename | `Submersion-v1.7.7.8064-Windows-Setup.exe`, identical in shape to the stable asset |
| Version number | Higher than stable, so it reads as newer and better |

Only the release title carries the word Beta ("Beta v1.7.7.8064").

The consequence is not cosmetic. Beta builds advance the schema ladder ahead of
stable (stable v1.7.6.7161 ships schema 175, main is at 191), and the ratchet in
`DatabaseService._openDatabase` means one accidental launch leaves the user
unable to open their dive log on the stable build they meant to be running. That
is #1568.

The in-app route to beta is well guarded by comparison: the channel picker in
`settings_page.dart` requires a second confirmation dialog that spells out the
database consequence. The direct-download route has no equivalent.

## Proposed

No app code required.

1. Publish beta releases with `--prerelease` in `.github/workflows/beta.yml` so
   GitHub stops labelling them Latest and shows the pre-release badge. Confirm
   first that this does not break the beta appcast, which is served from
   `beta-builds/releases/latest/download/appcast-beta.xml` and depends on
   `releases/latest` resolving. If a pre-release cannot be the `latest` target,
   pin the feed to a fixed tag or a separate asset host instead.
2. Prepend a short warning to the generated release body: these builds may
   upgrade the dive log database ahead of the stable release, and going back to
   stable afterwards will not open it again.
3. Add the same warning next to every beta-builds link in `README.md` and
   `docs/guide/installation.md`. The Linux package sections need it most, since
   they send stable users there by necessity.

## Files

- `.github/workflows/beta.yml` (the `gh release create` step)
- `README.md`
- `docs/guide/installation.md`

Related: #1568

ericgriffin:
Folding in the intent of #1592, which was attempted in #1600 and closed unmerged.

[...]

PR #1600, opened 2026-09-06
feat(release): give a beta build an in-app identity (#1592)

### LINKED ISSUE #1592: A beta build has no in-app identity, so direct-download users never know they are on one
## Summary

The release-channels design put build identity in the distribution feed only,
with no `-beta` suffix in binaries. That is right for the updater and wrong for
the person using the app: once a beta build is installed there is nothing in
About, the version string, or anywhere else in the UI that says so.

A user who reaches a beta build by direct download from `beta-builds` (linked
from the README and the install docs) therefore never learns they are on one,
and never sees the database warning that the in-app channel picker shows before
switching. They find out when a later stable build refuses to open their dive
log. See #1568.

Note this is distinct from `ReleaseChannel`, which is the user's update
preference, and from `UpdateChannel`, which is the compile-time distribution
channel (github/appstore/playstore). Neither says whether *this binary* came off
the beta train.

## Proposed

1. Stamp the build train into the artifact. A `--dart-define` set by
   `.github/workflows/beta.yml`, read the same way `UPDATE_CHANNEL` is read in
   `lib/features/auto_update/domain/entities/update_channel.dart`, defaulting to
   stable so nothing changes for existing builds.
2. Show it. A badge next to the version in Settings > About is the minimum. The
   version string itself stays untouched, since release tags and the appcast
   compare against it.
3. On first launch of a beta build, show the same warning the channel picker
   already shows (`settings_updates_betaDialogBody`): beta builds may upgrade
   the dive log database ahead of stable, going back will not downgrade it, and
   devices that sync together should share a channel. Users who arrive by direct
   download currently never see that text.
4. Include the train in bug reports and logs so an issue like #1568 is
   diagnosable from the report alone.

## Files

- `lib/features/auto_update/domain/entities/update_channel.dart` (pattern to follow)
- `lib/features/settings/presentation/pages/settings_page.dart` (About section)
- `.github/workflows/beta.yml` (the build step that would set the define)

Related: #1568

ericgriffin:
Closes #1592. Related: #1568.

## Problem

Beta binaries are configuration-identical to stable ones: beta-ness lives only in which repository publishes them. That is correct for the updater and wrong for the person using the app.

Someone who reaches a beta by direct download from `beta-builds`, which the README and `docs/guide/installation.md` link publicly, never learns they are on one. They never see the database warning the in-app channel picker shows before switching, and they find out when a later stable build refuses to open their dive log (#1568).

Neither existing concept answers "is this binary a beta?":

| | What it means | Why it does not answer the question |
| --- | --- | --- |
| `ReleaseChannel` | The user's update preference | A direct download leaves it on `stable` |
| `UpdateChannel` | Compile-time distribution store | Says how it was delivered, not which train built it |

## Change

**1. Stamp the train into the artifact.** `BuildTrain` / `BuildTrainConfig` read a `BUILD_TRAIN` dart-define exactly the way `UpdateChannelConfig` reads `UPDATE_CHANNEL`, defaulting to `stable` so nothing changes for a local `flutter run` or for any pipeline predating the define. Exposed as `buildTrainProvider`, because a compile-time const cannot otherwise be varied from a test.

**2. Show it.** The About badge now follows the *binary* rather than the *selected channel*. Those differ exactly where it matters: a direct-download beta user still has the stable preference, so reading the preference left that user with no badge at all. The selected channel is still shown, unchanged, on the update channel tile below.

**3. Warn on first launch.** A beta build shows the channel picker's own body text (`settings_updates_betaDialogBody`), so both routes onto the train read the same promise about backups and about sync partners sharing a channel. One `SharedPreferences` flag covers both routes, set when the picker is confirmed as well, so nobody reads the same text twice.

**4. Put it in the logs.** `LogEnvironment` carries the train in the session marker and the export header. Beta and stable share a version string, so nothing else in a pasted log separates them; this is what makes a report like #1568 diagnosable from the report alone.

The version string itself is deliberately untouched: release tags and the appcast compare against it.

## CI

`build-all.yml` takes a `build-train` input (default `stable`) and threads `BUILD_TRAIN` through all five `flutter build` steps, and via the environment through the two fastlane store lanes, which build both the App Store submission and the TestFlight beta from one lane. Only `beta.yml` passes `beta`; `release.yml` takes the default.

## Testing

- `build_train_test.dart`: parses every train, falls back to stable for null/empty/unknown, and reads stable when no define was passed.
- `beta_build_notice_test.dart`: the four-way decision table, plus the dialog rendering the picker's warning verbatim and closing on acknowledgement.
- `settings_page_test.dart`: a beta binary badges its version whatever the channel; a stable binary stays unbadged even on the beta channel (this replaces the old test that asserted the badge came from the channel, which is the behaviour being corrected); confirming the beta switch records the warning as read.
- `update_preferences_test.dart`: the new flag defaults to false and round-trips.
- `log_environment_test.dart`: the train appears in both the summary line and the export header, and `capture()` reads stable on an undefined host.

`flutter analyze` clean; the pre-push affected-test selection passed 631 tests.

### DIFF
--- a/.github/workflows/beta.yml
+++ b/.github/workflows/beta.yml
@@ -175,6 +175,11 @@ jobs:
       version-tag: ${{ needs.precheck.outputs.tag }}
       build-number: ${{ needs.precheck.outputs.build-number }}
       ref: ${{ needs.precheck.outputs.sha }}
+      # Stamps BUILD_TRAIN=beta into every artifact. The version string and the
+      # asset filenames stay identical to stable (tags and the appcast compare
+      # against them), so this define is the only thing that tells a user who
+      # direct-downloaded from beta-builds which train they are on (#1592).
+      build-train: beta
     secrets: inherit
 
   publish-beta:
--- a/.github/workflows/build-all.yml
+++ b/.github/workflows/build-all.yml
@@ -22,6 +22,14 @@ on:
         required: false
         type: string
         default: ''
+      build-train:
+        description: >-
+          Release train stamped into every artifact as the BUILD_TRAIN
+          dart-define, so the app can tell the user which train it came off.
+          Only beta.yml passes "beta"; everything else builds stable.
+        required: false
+        type: string
+        default: 'stable'
     outputs:
       eddsa-signature:
         description: 'Sparkle EdDSA signature of the macOS DMG'
@@ -46,6 +54,12 @@ on:
         required: false
         type: string
         default: ''
+      build-train:
+        description: 'Release train stamped into the artifact'
+        required: false
+        type: choice
+        options: [stable, beta]
+        default: stable
 
 permissions:
   contents: read
@@ -195,7 +209,7 @@ jobs:
           echo 'CODE_SIGN_IDENTITY = ' >> macos/Runner/Configs/Release.xcconfig
 
       - name: Build macOS release
-        run: flutter build macos --release --dart-define=UPDATE_CHANNEL=github --dart-define=GOOGLE_DRIVE_CLIENT_SECRET=${{ secrets.GOOGLE_DRIVE_CLIENT_SECRET }}
+        run: flutter build macos --release --dart-define=UPDATE_CHANNEL=github --dart-define=GOOGLE_DRIVE_CLIENT_SECRET=${{ secrets.GOOGLE_DRIVE_CLIENT_SECRET }} --dart-define=BUILD_TRAIN=${{ inputs.build-train }}
 
       # -- Developer ID DMG (for GitHub Release) --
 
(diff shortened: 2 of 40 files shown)


ericgriffin:
Closing unmerged. The approach in #1592 does not survive how promotion works.

`promote.yml` re-hosts the beta binaries bit-for-bit: it downloads the assets from the `beta-builds` release, checksum-verifies them, and re-uploads those exact files as the stable release under the same tag. The Sparkle EdDSA signature is copied out of the beta's `metadata.json`, so the DMG and EXE bytes have to be identical. Play promotes the same `version_code` and App Review gets the same TestFlight `build_number`.

So a compile-time `BUILD_TRAIN=beta` stamp survives promotion into the stable release. This branch would have badged every promoted stable build "(Beta)" in About, shown every stable user the beta database warning on first launch, and logged `beta train` in every bug report. That is wrong for a majority of installs and grows with each release.

The deeper problem is structural rather than a bug in the branch. "Is my schema ahead of stable?" is a fact that changes after the binary is frozen, so no compile-time constant can express it. `UPDATE_CHANNEL` works precisely because a store build's distribution channel never changes after the build.

It is also not the load-bearing fix. Stable `vX.Y.Z.N` is the promoted beta `vX.Y.Z.N`, same code and same schema version, so the bricking in #1568 is version ordering (build N upgraded the database, build M < N later refused it) and the train is incidental. #1591 addresses the same intent at the entry point, without any in-binary state, by marking beta releases pre-release and warning on the download page and in the docs.

Branch `claude/github-issue-1592-410cbb` is left in place if the CI plumbing (a `build-train` input threaded through `build-all.yml` and the two fastlane lanes) is wanted later.

PR #1607, opened 2026-09-06
feat(certifications): add FFESSM as a built-in agency

urbamax:
## What

Adds the **Fédération française d'études et de sports sous-marins (FFESSM)** as a built-in `CertificationAgency`. FFESSM is CMAS-affiliated but issues its own distinctly-named brevets across a modular system, so French divers currently file everything under `Other`, which discards the name.

A **slice of #690** — one agency, following the CMAS/BSAC/GUE pattern. Does not touch the table-backed refactor; that stays #690.

## Scope

The **scuba** cursus of the FFESSM *Manuel de Formation Technique* (août 2023), plus the scuba-diving activity commissions (biology, cave, underwater imaging). **47 `CertificationLevel` values.**

### Progression ladder — `_ffessmLadder` (what `primaryCertification()` ranks by)

Youth (`Plongeur de Bronze / Argent / Or`) → Niveaux (`N1`, `N2`, `N3`, `Guide de Palanquée (N4)`, `Directeur de Plongée (N5)`) → teaching (`Initiateur`, `Encadrant E2`, `MF1`, `MF2`).

### Modular qualifications — `_ffessmSpecialties` (not ladder rungs)

Held *alongside* a Niveau, not in place of one, so they sit outside the progression ladder. This list replaces the shared `specialties` pool for FFESSM in `specialtiesFor()`.

- **Aptitudes** — `PE` = *Plongeur Encadré* (supervised), `PA` = *Plongeur Autonome* (autonomous): `PE-12` `PE-40` `PE-60` · `PA-12` `PA-20` `PA-40`. Registered on their own (MFT *Généralités* p.3). `PE-20` is exactly N1 and `PA-60` is N3, so those two are not separately issued.
- **Tek** — nitrox (`Plongeur Nitrox`, `Plongeur Nitrox confirmé`, `Moniteur Nitrox confirmé`), trimix (`Plongeur Trimix élémentaire`, `Plongeur Trimix`, `Moniteur Trimix`), rebreather (`Plongeur recycleur SCR / CCR`, `Moniteur recycleur CCR`).
- **Safety** — `RIFA Plongée (RIFAP)` (required before N3 / N4 / Initiateur), `ANTEOR`.
- **Technical** — `Vêtement étanche`, `Sidemount de loisir`, `TIV`, `Formateur de TIV`.
[...]

Issue #553, opened 2026-07-10
Ability to add multiple certifications to buddy's or to myself


netdata-be:
For CMAS we have like 2 main certification levels:

- Main certification: Can be 1*D / 2*D / 4*D ....
- Extra certifications: Can be: Compressor operator, Nitrox diver, Advanced nitrox diver, Extended range diver

Any Idea how you would handle that?

ericgriffin:
There's already a rich feature set to define certifications for yourself in the "Certification" section of the app. I think the most flexible and future-proof way to handle buddies' multiple certs is to adopt the existing data schema and UI used for your own certs. That means you'll be able to store as much cert info for your buddies as you'd like, but all of the fields will be optional, so it could be as minimal as simply specifying the cert agency and level. We still need to identify a "primary" cert that can be used in the buddy picker as a badge, but I think using the highest cert in any of their agency's cert ladders will convey their general level when it's needed to quickly identify them.

ericgriffin:
Here's the logic that will be used to determine the "primary" cert (only used for display purposes where there's not space to list all of a buddy's certs - like in a UI picker):

  rank(cert) = index of cert.level in ladderFor(cert.agency),
               or -1 if level is null / not on that agency's ladder (i.e. a specialty)
  → pick the cert with the highest rank
  → tie-break: latest issueDate, then most recently updated
  → no certs, or all specialties with no ladder cert: still returns the
    top-ranked (or newest specialty) so the chip shows *something*; null only if empty

PR #1612, opened 2026-09-06
fix(buddies): show the certification’s "Name on the card" in the buddy list (#1303)

### LINKED ISSUE #1303: Use "Name on the card" under the name of buddies
Use "Name on the card" under the name of buddies.
Because when you use a certification "Other", it's just written "Other - Other" and you have to click on the profile to see the name of the certification...
### Summary
When a buddy holds a certification whose agency is set to **Other**, the buddy list only shows `Other - Other` as the certification line. The real certification name (the "Name on the card" field) stays hidden until you open the buddy's profile.

### Steps to reproduce
1. Add or edit a buddy.
2. Add a certification with **Agency: Other** / **Level: Other**, and fill in **Name on the card** (e.g. `FFESSM Niveau 2`).
3. Go back to the buddy list.

### Actual behavior
The certification line reads `Other - Other`. You have to open the buddy profile to find out which certification it actually is.

### Expected behavior
When agency/level are `Other`, show the **Name on the card** value under the buddy's name instead of `Other - Other`, so the list is readable at a glance.

### Related requests (lower priority, could be split out)
- Custom certification agencies / levels, so users can define their own instead of falling back to `Other`.
- Built-in FFESSM certifications (Niveau 1/2/3, Plongeur Autonome/Encadre 20/40/60, etc.).

### Environment
- Submersion v1.7.5.6772, macOS

Image
And by the way, custom certification would be really great... Or please add FFESSM certifications...

Image

v1.7.5.6772 on macOS

urbamax:
## What

When a buddy's primary certification has **Agency = Other**, the buddy list, pickers and detail header show `Other · Other` (or just `Other`). The real certification — the **Name on the card** field, e.g. `FFESSM Niveau 2` — is only visible after opening the buddy's profile.

Fixes #1303 (the display half; the "built-in FFESSM / custom agencies" half is #690 / #1607).

## Changes

- `Buddy` gains a derived `certificationTitle`, hydrated in `BuddyRepository._withPrimaryCerts` from `certificationTitle(primary)` — the same helper the certification surfaces already use: the stored custom name when it adds anything over the agency/level pair, otherwise the derived title.
- New `Buddy.certificationLine` getter: the title, then the agency — **unless** the agency is `Other` (whose "Other" carries no information) or is already part of the title. Wired into the card tile, dense tile, summary preview, buddy picker and the detail header. `Buddy.displayName` now prefers the title too.
- The configurable **"Cert Level" table column** (`BuddyField.certificationLevel`) is deliberately left as the level enum — that column is explicitly the level.

## No schema / sync change

`certificationTitle` is derived at read time (like `certificationLevel` / `certificationAgency` since #553) and is `null` on the raw row. Nothing persisted, no migration.

## Tests

New `test/features/buddies/domain/entities/buddy_test.dart` covers `certificationLine` (Other-agency custom name → just the name; real agency → kept; agency already in title → not repeated) and `displayName`. `flutter analyze` (whole project) + `dart format` clean; buddy + certification suites green (543 tests).


urbamax:
6167e74c7 — closed the Codecov patch gap: widget tests for the two `certificationLine` render sites — the `buddy_summary_widget` list-tile subtitle, and the embedded `buddy_detail` header line. The getter's branches were already covered in `buddy_test.dart`.

urbamax:
Both Copilot threads addressed in 98a58880:

- **Case/whitespace-sensitive agency match** — `certificationLine` now compares the title against the agency name after stripping case and non-alphanumerics, so `"Padi Rescue Diver"` / `"PADI - Rescue Diver"` no longer get `· PADI` appended. New test cases in `buddy_test.dart`.
- **Dense tile not using `certificationLine`** — `dense_buddy_list_tile` now renders `buddy.certificationLine` (was `certificationTitle ?? certificationLevel.displayName`), matching the other buddy surfaces and the PR description. It also picks up the agency-only case now.

Issue #1523, opened 2026-09-03
Suunto Nautic: dive-computer events overlap computed ones, and lose their exact labels


urbamax:
Context: testing a local build with Suunto Nautic support (libdivecomputer driver from libdivecomputer/libdivecomputer#73 + the event-mapping fix in deepsealabs/libdivecomputer-upstream#1). A ready-to-build fork with all of it: https://github.com/urbamax/submersion/tree/suunto-nautic (see `SUUNTO_NAUTIC.md`). The Nautic reports its own dive events. Two problems on the profile:

> **Status (fork):** both parts are implemented on the fork.
> - Part 1 — "Computed events" legend toggle (`b0e35a1`): session-only, no `AppSettings` field, off by default when the dive has imported events.
> - Part 2 — exact Suunto labels (`dd05a22`): `suunto_nautic_event_labels.dart` decodes `event.value` = `(sub-group << 8) | type` to the app's wording ("Ascent rate alarm", "High ppO2 alarm" vs "Low ppO2 alarm", "Deco stop reached" vs "Deco stop broken", …) into `ProfileEvent.description`; chart markers use it via a new `markerLabel` getter. Import + reparse, Suunto Nautic / Ocean only. The ascent-rate markers also show the rate at that moment in the diver's unit ("Ascent rate alarm · 14.0 m/min").
>
> Still needs a **driver** change: the two events the Suunto app shows but libdivecomputer's `suunto_nautic_map_event` maps to `SAMPLE_EVENT_NONE` (Warning type 20 "low no-deco time", State type 19 "became a deco dive") are never emitted, so nothing app-side can label them. They can instead be surfaced as `EventSource.computed` markers from the app's own NDL / deco calc.

## 1. Computed events overlap the computer's own events

The app's computed ascent-rate warning sits ~1 s away from the watch's own ascent alarm, computed safety-stop markers overlap the watch's, etc. They stack and become unreadable.

**Cause:** `mergeEvents()` in `lib/features/dive_log/domain/services/profile_event_mapper.dart` dedupes computed vs imported events only on an exact `(timestamp, eventType)` key. A computed warning at `t=1081` and the computer's at `t=1082` are different keys, so both survive.

[...]

PR #1616, opened 2026-09-06
feat(dive-profile): separate "Computed events" legend toggle

urbamax:
Part 1 of #1523 — a standalone profile-UI problem, not Suunto-specific.

## The problem

On a dive downloaded from a computer that reports its own events, the app's auto-detected ascent-rate / safety-stop markers land a second or two away from the computer's own and stack into an unreadable smear. `mergeEvents()` only dedupes computed vs. imported on an exact `(timestamp, eventType)` key, so a computed warning at t=1081 and the computer's at t=1082 both survive.

## This change

Add a **"Computed events"** entry next to "Events" in the inline legend and the chart-options dialog. It shows/hides only `EventSource.computed` markers; the computer's own events and user bookmarks stay on the plain "Events" toggle. The row only appears when the dive carries both kinds of event (`ProfileLegendConfig.hasSplitEvents`).

Default: hidden when the dive carries the computer's own (imported) events, shown otherwise — so a computer download shows just the computer's events, and a manually logged / file-imported dive still shows the app's analysis. An explicit user toggle sticks for the session. Session-only, no persisted setting, matching the other legend toggles.

## Scope / testing

- ~215 lines, additive. No change to event merging itself or to any default a dive without split events sees.
- New: 6 provider tests (`profile_legend_provider_test.dart`) + 3 widget tests (`dive_profile_chart_computed_events_test.dart`) covering seed-from-dive, user-toggle-wins, and the plain-Events interaction.
- New ARB key `diveLog_legend_label_computedEvents` in all 11 locales; `arb_parity_test` green; `flutter gen-l10n` regenerates cleanly; `test/features/dive_log/presentation/widgets/` (1182 tests) green.

Surfaced with the Suunto Nautic (#1523) but any computer whose events overlap the app's computed ones benefits. Part 2 of #1523 (exact per-vendor event labels) stays on the fork, blocked on the driver upstream.


urbamax:
Addressed in 8eb52ac.

I went a slightly different route than "make the seed synchronous". Seeding the provider synchronously from `initState` is fragile here: `profileLegendProvider` is `autoDispose` and is read by the legend widgets, so a synchronous `state = …` while another widget is building would throw. Instead:

- The chart derives `_showComputedEvents` on its **own first build**, mirroring `seedComputedEventsVisibility`'s rule (`!diveHasImportedEvents`) while `computedEventsFollowsDive` is true (the user hasn't touched the toggle). The first paint is already correct, and stays correct when switching between dives.
- The post-frame `seedComputedEventsVisibility` hop stays, but now only to keep the shared provider — and the legend checkbox — in step. No visual effect.

Also reworded the `showComputedEvents` doc comment: it's a sub-filter of `showEvents`, not independent of it (event markers only draw when `showEvents` is on).

New coverage: `dive_profile_chart_computed_events_test.dart` now pumps a single frame and asserts no computed line is drawn on an imported-events dive; `profile_legend_provider_test.dart` covers `computedEventsFollowsDive`.

ericgriffin:
I think this is no longer relevant as implemented now that the legend has been made non-interactive (can't click to deselect). This should probably be implemented in the chart options dropdown similar to the DC/Calc toggle for dive computer metrics.

ericgriffin (changes requested):
See prior comment

urbamax:
@ericgriffin — rebased on current main (l10n conflicts resolved).

On "implement it in the chart options dropdown": the toggle **is** already in the chart-options dialog — `_buildOverlaySection` in `chart_options_dialog.dart`, gated on `config.hasSplitEvents` (only appears when the dive carries both the computer's own events and the app's auto-detected ones). It's a plain on/off `buildToggleItem` there, next to Events / Temperature / Pressure.

The only legend-side change left is one **read-only** row in `active_legend_entries.dart` — that widget just lists what's currently drawn (with the matching colour swatch); it has no tap handler, so the non-interactive-legend change doesn't affect it. Happy to drop that row too if you'd rather the legend not mention computed events at all.

Open question for you: do you want it kept as a simple show/hide toggle, or turned into a `buildToggleWithSource` on the **Events** row itself — a "Computer / Calculated / Both" segmented control like the NDL/TTS/deco-stop source rows? The events case is a bit different from those (you often want to see both marker sets at once, so "Both" would be the default), but I can build it that way if it's more consistent with the DC/Calc pattern you had in mind.

urbamax:
To be concrete about where this stands: the "Computed events" control is a plain on/off toggle in the chart-options dropdown (`_buildOverlaySection` → `buildToggleItem`), right next to Events / Temperature / Pressure, and only shows when the dive actually carries both event kinds. That's exactly "in the chart options dropdown" — nothing legend-interactive here.

The one remaining legend touch is a single **read-only** entry in `active_legend_entries.dart`: when computed markers are being drawn, the legend names them with the matching amber swatch (same as every other drawn metric). No tap target, so the non-interactive-legend change doesn't reach it.

I'd like to keep it as this simple show/hide toggle rather than a Computer/Calculated segmented control — the two event sets are different marker types you usually want visible together, not an either/or. Could you re-review with that in mind? If the read-only legend row is the sticking point I'll drop it, just say the word.

urbamax:
@ericgriffin when you get a moment — this one's ready for another look (the toggle is in the chart-options dropdown as you asked; details two comments up).

While you're in here: the rest of the v1.7.8 batch is green and clean too — #1613, #1615, #1620, #1624, #1647 have no open threads, and #1629 (the enum-label localization — fixes the untranslated Weight Belt / Integrated Weights / Salt Water strings in the 8090 beta) and #1623 both just went green after the l10n regen. No rush, just flagging they're unblocked.

ericgriffin (approved):
👍

PR #1627, opened 2026-09-06
Resync file-imported dives from their original file (#478)

### LINKED ISSUE #478: Parser fixes never reach already-imported dives (no re-parse path for file imports)
## Problem

Import parses a file once and persists the derived rows (`dive_tanks`, `gas_switches`, profile samples). Parser improvements therefore never reach dives that were imported before the fix shipped.

Concrete case: #404 / PR #471 adds gas-switch and multi-gas import for Shearwater UDDF and Garmin FIT. Users who already imported affected dives still see single-gas profiles, because:

- The re-parse machinery (`ReparseService`) only operates on `dive_data_sources` rows with `rawData`, which only libdivecomputer downloads populate. File imports (universal import) store no raw payload — there is nothing to re-parse from.
- Re-importing the same file offers only *Skip*, *Import as New*, and *Consolidate as additional computer*. None updates the existing dive. *Import as New* creates a near-duplicate that the deco engine chains as a repetitive dive (zero surface interval, inherited tissue load — misleading TTS/SurfGF/heatmap).
- The only workaround is delete + re-import, which destroys manual edits (notes, buddies, site, rating, equipment links).

The same wall will be hit by every future parser fix (e.g. Shearwater `calculatedpo2` is still unparsed).

## Solution paths

**A. Store imported file bytes in `dive_data_sources.rawData`** (column exists; DC downloads already use it). Makes every future parser fix retroactively applicable via the existing `ReparseService` — but only for dives imported after this ships. Open questions: DB growth (FIT ~200 KB, UDDF up to MBs per dive), whether `rawData` should sync or stay local-only.

**B. Add an "Update dive data" duplicate resolution to the import wizard.** Keeps the dive row and all manual edits; rebuilds computer-derived data (tanks, profile, gas switches, tank pressure profiles, `dive_data_sources`) from the fresh parse. This is the only option that helps dives imported in the past, and it mirrors the existing replace-source re-download flow for dive computers. Care needed with cascades: gas switches and pressure profiles are keyed to tank rows being replaced (see the #276 SAC regression for prior art).

**C. One-time backfill migration — infeasible.** The gas data exists only in the source files, which were never stored.

**D. Release-note workaround only** (delete + re-import): zero cost, loses manual edits, and repeats for every future parser fix.

## Recommendation

Implement **B** first (lossless upgrade path for existing users as soon as a parser fix ships), then **A** (so this is the last time users have to re-pick files at all).

Refs: #404, #471

readme42:
Closes #478.

Import parsed a file once, persisted the derived rows and threw the file away, so every later parser fix was invisible to dives already imported. The only workaround was delete and re-import, which destroyed notes, buddies, site, rating and equipment links along with the stale data. This branch keeps the file and adds a "Resync from original file" action that re-parses it and rewrites the computer-derived data onto the existing dive.

Resync button: 
image

Stats: 
image

## How it works

The file is stored in the database, not on disk: a new `imported_files` table (schema v208) holds the original bytes, zlib-compressed at rest behind the same self-describing header `dive_data_sources.raw_data` uses. The row id is the sha256 of the bytes, so one import run stores one row however many dives came out of the file, and re-importing the same file reuses the row it already has. Each dive's source row points at it through `imported_file_id`. The table is a synced entity with its own `hlc`, so the file reaches the diver's other devices and is covered by backups, which is what put it in the database rather than a folder (see the discussion below).

`DiveResyncOrchestrator` loads the stored bytes, picks the parser the recorded format names, finds the right dive inside the file with the existing `DiveMatcher` — a stored file may be a whole logbook — and hands the match to `DiveReimportService`. Every way it can decline is a typed failure the page renders as a localized sentence in all twelve locales: the format cannot be re-parsed, no file was stored, the row has not reached this device, or nothing in the file matches the dive any more.

## What a resync writes

It behaves like the dive-computer re-parse, deliberately and field for field. The dive row gets the parse's runtime, entry and exit time, bottom time, max and average depth, dive mode, CNS, OTU, deco model, conservatism and gradient factors; water temperature and GPS are left alone when the parse has none, exactly as `ReparseService` leaves them. Profile samples, gas switches, tank pressures and profile events are replaced. The source row's snapshot is rewritten including nulls, so a parser fix that stops emitting a value clears the stale one instead of leaving it to contradict the rest of the dive.

That means a hand correction to a computer-reported value — a max depth someone typed over — is overwritten by a resync, the same way the computer re-parse overwrites it today. Genuinely diver-authored data is untouched: notes, buddy, dive master, rating, name, site, dive center, trip, course, equipment, tags, weights, sightings, dive types, the statistics exclusions and the favourite flag.

Writes are scoped to what the source owns. `sourceOwnsProfileStrand` moves out of `ReparseService` so both paths share one rule, and on a dive that carries a second source the other source's profile, pressures and gas switches are preserved rather than destroyed — the action then reports that the profile was preserved instead of claiming a full update. `dive_tanks` rows are carried over rather than deleted and recreated, because `gas_switches` and `tank_pressure_series` both cascade off `dive_tanks.id` and issue #276 is what that costs when it goes wrong; tanks match in two passes, an explicit `order` claiming its row first and the rest pairing by list position, because `order` is not an identity every parser supplies.

Every path that removes a source row reclaims the bytes behind it: deleting a dive or a batch, deleting a diver with their dives, replacing a computer's source, and applying a remote tombstone for either a dive or a source row. The refcount is an anti-join rather than a counter the callers carry, so the answer is the same whatever did the deleting. Reclamation stays local on purpose: a device never tells its peers to drop a file, because one that is a sync behind still has the dive and would be told to throw away bytes it needs. Storage Usage reports what the rows occupy, under "Raw imported data".

## Scope

CSV needs a stored field mapping and gets none; `macdiveSqlite`/`sqlite` are multi-file DB imports; `unknown` has no parser. Those store no file and are never offered a resync. A file carrying its own `<source>` entries — which is what Submersion's own exports produce — takes the restored-rows path and stores nothing either. A dive imported before this ships has no stored file, so it gets no resync: this closes the hole going forward rather than retroactively. Subsurface's placeholder-cylinder `tankIndex` skew is broken identically on the import path, so a resync is no lossier than the import it replays; both are wrong and neither is fixed here.

Follow-ups discussed in the thread and deliberately left out: a bulk reparse of the whole database, a management page for stored files, the same raw storage for the cloud import sources, matching an improved third-party export against existing dives, and re-parsing automatically after an app update.

## Testing

`dive_import`, `dive_computer`, `core/services/sync`, `core/database`, `import_wizard`, the storage inventory and the architecture guards all pass locally, and `flutter analyze lib test` is clean. Known failures unrelated to this branch, all reproducible on `main`: the deco-axis filter test, `dive_computer_repository_impl_test`'s relink case when its whole file runs, the PDF logbook template test, three photo-folder-resolution tests, and `provider_tick_scanner_test`, which fails only on Windows because its fixture builds a temp path with mixed separators.

To try it by hand: import a foreign UDDF or FIT file, edit the dive, then use "Resync from original file" in the dive's overflow menu and check the computer data is re-derived while the edits stand.


ericgriffin:
@readme42 what are your thoughts on potentially storing the raw file data (compressed of course) alongside the dive in the database, similar to the way we store raw dive computer data? Only the data pertaining to the dive would be isolated. Doing it this way, the source import data would be able to be synced across devices, and it would also be included in backups. My concern with keeping it in a local directory is that it's not particularly durable on mobile devices where the application directory can be deleted if the app is uninstalled, and it's not synced. I'm not sure what the storage cost would be, but I can't imagine it would make too much of an impact, and this type of data compresses very well.

readme42:
Thanks @ericgriffin — that was actually my first instinct too, and `rawData` is where I started.

What changed my mind was the multi-dive case: logbook exports and SQLite databases hold many dives in one file. `rawData` is per-dive, so I'd have had to either store the same file once per dive, or cut out "the part for this dive" — and a logbook can't be cut that way, because sites, gas mixes and buddies live in shared sections that the dives only reference. Duplicating a 400-dive export 400 times is what pushed me to the folder instead.

That said, I'd prefer the database too, for exactly the reasons you give — and for anyone running with database encryption enabled, the files would be covered by it, which they never are sitting in the documents folder.

So I think the answer is a new table rather than `rawData`: one row per file, keyed by content hash, referenced by the source rows that came from it. That keeps the original bytes intact, stores each file once no matter how many dives it holds, and turns the cleanup I built across four delete paths into a single refcount query.

Backups come with being in the database at all, so that's settled either way. What syncing the table adds is exactly the second-device case you're pointing at, and I'd vote for syncing it — a dive that reaches another device should bring its source file with it. One row per file keeps the payload far smaller than the per-dive version would have been, though it's worth keeping an eye on large migrations.

So if you are in, i would adjust to a new table instead of a documents folder.

readme42:
> [Play the flyover — 6 chapters, 23:15](https://flyovers.dev/flyover/c6986e09-180c-4027-85ba-ff763f45fcc2)
> 
> **[Watch the flyover →](https://flyovers.dev/flyover/c6986e09-180c-4027-85ba-ff763f45fcc2)** · 23:15 · 6 chapters · 68 files
> 
> An animated walkthrough of this diff: context, the problem, the change, implementation care, code walkthrough, and impact.
> 
> [f231141](https://github.com/submersion-app/submersion/commit/f231141a62e75e397bb6fff5ee7b15ef70a279c4) · updated automatically by flyovers

A bit verbose on explaining test cases :D

ericgriffin:
> Thanks @ericgriffin — that was actually my first instinct too, and `rawData` is where I started.
> 
> What changed my mind was the multi-dive case: logbook exports and SQLite databases hold many dives in one file. `rawData` is per-dive, so I'd have had to either store the same file once per dive, or cut out "the part for this dive" — and a logbook can't be cut that way, because sites, gas mixes and buddies live in shared sections that the dives only reference. Duplicating a 400-dive export 400 times is what pushed me to the folder instead.
> 
> That said, I'd prefer the database too, for exactly the reasons you give — and for anyone running with database encryption enabled, the files would be covered by it, which they never are sitting in the documents folder.
> 
> So I think the answer is a new table rather than `rawData`: one row per file, keyed by content hash, referenced by the source rows that came from it. That keeps the original bytes intact, stores each file once no matter how many dives it holds, and turns the cleanup I built across four delete paths into a single refcount query.
> 
> Backups come with being in the database at all, so that's settled either way. What syncing the table adds is exactly the second-device case you're pointing at, and I'd vote for syncing it — a dive that reaches another device should bring its source file with it. One row per file keeps the payload far smaller than the per-dive version would have been, though it's worth keeping an eye on large migrations.
> 
> So if you are in, i would adjust to a new table instead of a documents folder.

@readme42 yep, sounds good to me. New synced table in the DB (w/compressed file data) keeps it simple and fool-proof. Thanks!

ericgriffin:
> > [Play the flyover — 6 chapters, 23:15](https://flyovers.dev/flyover/c6986e09-180c-4027-85ba-ff763f45fcc2)
> > **[Watch the flyover →](https://flyovers.dev/flyover/c6986e09-180c-4027-85ba-ff763f45fcc2)** · 23:15 · 6 chapters · 68 files
> > An animated walkthrough of this diff: context, the problem, the change, implementation care, code walkthrough, and impact.
> > [f231141](https://github.com/submersion-app/submersion/commit/f231141a62e75e397bb6fff5ee7b15ef70a279c4) · updated automatically by flyovers
> 
> A bit verbose on explaining test cases :D

This also opens the door to allowing someone to reparse/repopulate their entire database (dc raw + file) as a bulk operation, and a management page to allow one to delete/replace/remap previously imported files. We should consider doing the same raw data storage for the cloud import data sources as well.

We could also consider the case where another application's exporter gets an enhancement/bug-fix, and the user wants to re-import the new (better) exported dives, but match them up with existing dives in Submersion. Maybe that's overkill, but if it's not, it could be a follow-up enhancement.

readme42:
> ort the new (better) exported dives, but match them up with existing dives in Submersion. Maybe that's overkill, but if it's not, it could be a follow-up

Yeah, there are many opportunities how we can utilize the store data - at least as long as its a somehow imported dive. A manual entry cant get better ;)

We could also, when we improved our own parser or importer, - eg after a new release was installed - reparse existing dives automatically to bring new feature/fixes without the need for the user to manually reparse. 

Tons of possibilities - the sky is the limit ;) I would just not consider them as in scope of this PR but for future improvements.

lib/features/dive_import/data/services/dive_resync_orchestrator.dart
```diff
@@ -0,0 +1,133 @@
...
+    // `duration`, so `candidate['duration']` is always null there and the
+    // match rides on time 0.50 + depth 0.30 = 0.80 with no margin.
+    Map<String, dynamic>? best;
+    var bestScore = 0.0;
+    for (final candidate in candidates) {
+      final candidateTime = candidate['dateTime'] as DateTime?;
+      if (candidateTime == null) continue;
+      final candidateDepth = (candidate['maxDepth'] as num?)?.toDouble() ?? 0.0;
+      final candidateDuration =
+          (candidate['duration'] as Duration?)?.inSeconds ?? 0;
```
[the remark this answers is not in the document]
readme42: Good catch, fixed. The score now reads `candidate['duration'] ?? candidate['runtime']`, so UDDF and the other runtime-only payloads keep the duration weight instead of silently scoring zero on it — which mattered most exactly when a parser fix shifts depth and the match needs the margin.

test/core/services/storage/storage_inventory_test.dart
```diff
@@ -51,19 +52,23 @@ void main() {
       networkImageDirectory:
           networkImageDirectory ??
           () async => Directory(p.join(temporary.path, 'libCachedImageData')),
+      importedFilesDirectory:
+          importedFilesDirectory ??
+          () async => Directory(p.join(documents.path, 'imported')),
     );
```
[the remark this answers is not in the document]
readme42: Resolved by the storage redesign rather than by fixing the path: there is no documents folder any more. The bytes now live in the synced `imported_files` table, compressed, one row per file, and the inventory measures those rows — so the test injects a byte count instead of a directory.

lib/features/dive_import/data/services/dive_resync_orchestrator.dart
```diff
@@ -0,0 +1,137 @@
...
+          0;
+      final score = _matcher.calculateMatchScore(
+        wearableStartTime: candidateTime,
+        wearableMaxDepth: candidateDepth,
+        wearableDurationSeconds: candidateDuration,
+        existingStartTime: DateTime.fromMillisecondsSinceEpoch(
+          dive.diveDateTime,
+        ),
+        existingMaxDepth: dive.maxDepth ?? 0.0,
+        existingDurationSeconds: dive.bottomTime ?? 0,
```
ericgriffin: This compares the candidate's total dive time (`duration ?? runtime`) against the stored dive's `bottomTime`, which the importer derives from the profile without the ascent and stops. On most dives the two differ by 5 to 10 minutes, so the duration score is mostly lost.

That matters in exactly the case the comment above describes. Example: a UDDF dive with runtime 50 min and a profile-derived bottomTime of 42 min. A parser fix corrects an ft/m depth bug, so maxDepth moves by more than 20% and the depth score is 0. Time gives 0.5, the 8 min duration diff gives about 0.06, total about 0.51. That is under 0.70, so the resync returns `noMatchingDive` for the fix it exists to deliver.

The importer fills `dives.runtime` from `runtime ?? duration`, so comparing `candidate['runtime'] ?? candidate['duration']` against `dive.runtime` would be like for like.

lib/features/dive_import/data/services/dive_reimport_service.dart
```diff
@@ -0,0 +1,620 @@
...
+  }) async {
+    final maxDepth = _asDouble(diveData['maxDepth']);
+    final avgDepth = _asDouble(diveData['avgDepth']);
+    final waterTemp = _asDouble(diveData['waterTemp']);
+    final bottomTime = _deriveBottomTimeSeconds(diveData);
+    final decoAlgorithm = diveData['decoAlgorithm'];
+    final gfLow = diveData['gradientFactorLow'];
+    final gfHigh = diveData['gradientFactorHigh'];
+
+    await (db.update(db.dives)..where((t) => t.id.equals(diveId))).write(
```
ericgriffin, line 276: `bottomTime` is recomputed here and `_replaceProfile` swaps the samples, but `dives.runtime` and `dives.exitTime` are never rewritten. The importer sets both from the same parse (`runtime = diveData['runtime'] ?? duration`, `exitTime = dateTime + runtime`).

After a parser fix that changes the profile length (the OSTC profile-length class of bug), the profile and bottom time describe a 45 min dive while the header, list, surface-interval and overlap logic still read the old 60 min runtime and exit time. Should `runtime` and `exitTime` join the allowlist, with the same fallback the importer uses?

ericgriffin (commented):
Two correctness issues on the resync path, inline below.

Separately, @readme42, could you update the PR description? The "How it works" section says the picked bytes are copied into `imported/` under the documents directory and that each `dive_data_sources` row records a relative path. The code stores the file as a zlib-compressed blob in the new `imported_files` table, keyed by sha256, and the source row records `imported_file_id`. The comment above the `importedFiles` entry in `SyncRepository` (the one about files "adopted out of the pre-database `imported/` folder" by a migration) looks like it is left over from the same earlier design, since the v208 rung has no adoption step.

lib/features/dive_import/data/services/dive_resync_orchestrator.dart
readme42: You're right, and thanks for the worked example — it made the problem obvious. The candidate's total time was scored against `dive.bottomTime`, which the importer derives from the profile without ascent and stops, so the duration weight was mostly thrown away. It now compares `runtime ?? duration` against `dive.runtime` (falling back to bottomTime only when runtime is null), which is like for like, and your UDDF example is now a test: runtime 50 min against a stored 50 min runtime, depth scoring zero after an ft/m fix, and the match clears the threshold instead of returning `noMatchingDive`.

lib/features/dive_import/data/services/dive_reimport_service.dart
readme42: Yes, they should, and they do now. The dive row mirrors `ReparseService._updateDiveRow` field for field: runtime and exitTime with the importer's own derivation (`runtime ?? duration`, exit = dateTime + runtime), plus diveDateTime, entryTime, bottomTime, CNS, OTU, deco model, conservatism and gradient factors; water temperature and GPS stay untouched when the parse has none, exactly as the re-parse leaves them. One deliberate difference: dive mode is only written when the parse states one. UDDF, FIT and MacDive state none, and a re-parse always writes a mode it actually read, so defaulting to OC would have turned every resynced CCR dive into an open-circuit one.

lib/features/dive_import/data/services/dive_reimport_service.dart
```diff
@@ -0,0 +1,620 @@
...
+    required DiveDataSourcesData? primary,
+    required Map<String, dynamic> diveData,
+    required DateTime now,
+  }) async {
+    if (primary == null) return;
+
+    final maxDepth = _asDouble(diveData['maxDepth']);
+    final avgDepth = _asDouble(diveData['avgDepth']);
+    final duration = _asDurationSeconds(diveData['duration']);
+    final waterTemp = _asDouble(diveData['waterTemp']);
```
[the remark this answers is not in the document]
readme42: Correct, fixed: the snapshot's `duration` now comes from `_deriveBottomTimeSeconds(diveData)`, the same derivation the import path uses when it stamps that column, so a UDDF resync no longer leaves the Sources panel on the old value.

readme42:
Thanks — all of it was worth raising, and the resync now follows one rule rather than inventing its own: it behaves like the dive-computer re-parse. Working through the suppressed findings as well:

Fixed: the dive row now mirrors `ReparseService` (runtime, exit time, clock, CNS, OTU, deco model, conservatism, GFs; water temperature and GPS left alone when the parse has none). Profile events are replaced under the same ownership gate a re-parse uses, with the importer's event conversion extracted into one shared mapper instead of a second copy. The source snapshot is written including nulls, so a parser fix that stops emitting a value clears the stale one. The snapshot's duration follows the importer's derivation. `_replaceTankPressures` now scopes its delete to the tanks that actually produced fresh samples, so a parsed tank without pressures keeps its history. A remote source-row tombstone runs the reclaimer, and `ImportedFileRepository` and `ImportedFileReclaimer` bind their `SyncRepository` to the injected database. `diveHasImportedFileProvider` also ticks on `imported_files`, so the action appears when the blob arrives after the reference it belongs to. The leftover comment about files adopted out of the old `imported/` folder is gone, along with the folder itself.

Two I found while verifying, not from the review: events were read only under `events`, but the UDDF path fills `profileEvents`, so the commonest resync replaced nothing while reporting success; and dive mode defaulted to open circuit when the parse stated none, which would have turned resynced CCR dives into OC ones.

Deliberately not done: per-field provenance for manual corrections. A resync overwrites computer-reported values, including a max depth someone typed over, exactly as a computer re-parse does today — the service's doc comment now says that instead of promising otherwise, and genuinely diver-authored data (notes, buddy, site, rating, tags, trip, equipment, weights, sightings) stays untouched. Also left: the `.valueOrNull` reload-flicker note, since the raw-data button beside it uses the identical pattern; if it flickers, it flickers for both and belongs in its own change.

PR description updated — it described the folder design, which no longer ships.

PR #1630, opened 2026-09-06
suunto cloud: import the full dive-event set, not just three types

urbamax:
## What

The Suunto Cloud import (`suunto_cloud/`, a Dart port of Subsurface's `import-suunto-json.cpp`) only recognised three events: gas switches, "At Safety Stop", and the ascent-speed alarm. A real dive's Suunto app view shows a dozen; the rest were silently dropped by `_collectEvents`.

## Change

The `sml` JSON carries every event per sample under `DiveEvents.<Alarm|Warning|Notify|State|Ooam>.Type` as a descriptor string. New `suunto_cloud_event_map.dart` maps `(subgroup, Type)` -> a libdivecomputer-style event type string (built from Suunto's own descriptor enums, `com.stt.android.domain.sml`), so the existing shared `_mapEventTypeString` pipeline turns them into `ProfileEventType` markers:

| Suunto event | ProfileEventType |
|---|---|
| Alarm/Warning/Notify tank-pressure + gas-time | `lowGas` |
| Alarm ascent-speed | `ascentRateWarning` |
| Alarm/Warning CNS% / OTU | `cnsWarning` / `cnsCritical` |
| Alarm/Ooam deco-ceiling / deep-stop broken | `decoViolation` |
| State "At Deco/Deep/Safety Stop" | `decoStopStart` / `safetyStopStart` |
| Alarm "Safety Stop Broken" | `missedStop` |

- `_collectEvents` now walks every sub-group on **both** container shapes (the `DiveEvents` object and the `Events[]` array), rising-edge only -- a begin the watch holds across several samples imports once, a re-trigger imports again.
- It also stamps `DownloadedEvent.value` with the watch's native `(subgroup << 8 | type)` code. Inert today; ready for a consumer that renders the exact Suunto wording.
- `_mapEventTypeString` / `_eventSeverity` gain `cnsWarning` / `cnsCritical` / `missedStop` passthrough (Suunto-cloud-only strings; no effect on any dive-computer download path).

**Deliberately not imported** (unchanged): battery, sidemount, recovery-time, bearings/stopwatch, and the predictive "... Ahead" states. Two informational states ("low no-deco time", "became a deco dive") are noted as a follow-up -- they have no `ProfileEventType` yet.

## Testing

`flutter analyze` clean. New `suunto_cloud_event_map_test.dart`; `suunto_dive_parser_test.dart` gains a "full dive-event extraction" group (the newly-imported events, the native-code stamp, the held-begin / re-trigger edge behaviour). `test/core/services/suunto_cloud`, `test/features/dive_log/data` and the import-wizard suites are green.

### DIFF
--- a/lib/core/services/suunto_cloud/suunto_cloud_event_map.dart
+++ b/lib/core/services/suunto_cloud/suunto_cloud_event_map.dart
@@ -1,0 +1,87 @@
+/// Maps a Suunto cloud/app `sml` dive-event -- `DiveEvents.<Subgroup>.Type` as
+/// a descriptor string -- to a libdivecomputer-style event-type string
+/// (consumed by the dive-computer repository's `_mapEventTypeString`) plus the
+/// watch's native `(sub-group << 8) | type` code.
+///
+/// The string ⇄ code correspondence is Suunto's own descriptor library, as
+/// decoded from a Nautic's `/Logbook/byId/<id>/Descriptors` (5 enums under
+/// `com.stt.android.domain.sml`: `AlarmMarkType`, `WarningMarkType`,
+/// `NotifyMarkType`, `StateMarkType`, `OoamMarkType`).
+///
+/// A cloud dive-event carries `Active: true` on its begin edge and, on some
+/// generations, `Active: false` on the end edge; only the begin edge is a
+/// marker. The older `DiveEvents` *object* shape has no `Active` field -- treat
+/// it as a begin.
+///
+/// The native code is stored on `DownloadedEvent.value` and is inert today; a
+/// consumer that decodes it (`suuntoNauticEventLabel`, currently on the
+/// Suunto Nautic fork) can render the exact Suunto wording.
+library;
+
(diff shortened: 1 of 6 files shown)


urbamax:
Copilot's native-code point is fixed in eb25143.

The `Notify` / `Safety Stop` alias was stamping the code as `(_state << 8) | 37` — `_state` (0x1B) under a `Notify` key, which breaks the `(sub-group << 8) | type` contract. It's a legacy Subsurface string with no place in the Nautic descriptor (the Nautic announces the stop through `State` "At Safety Stop"), so there is no Notify type number to give it. `SuuntoCloudEvent.nativeCode` is now nullable and this entry carries no code; `suunto_cloud_event_map_test.dart` asserts that, and that every real entry's high byte matches its sub-group.

f6e23cb adds an `importProfile` test for the `cnsWarning` / `cnsCritical` / `missedStop` passthrough, closing the Codecov patch gap.

Issue #1642, opened 2026-09-07
Bottom time can come out longer than the dive's own total duration


trustthegoose:
# Bottom time can come out longer than the dive's own total duration

`BottomTimeCalculator.secondsFromSamples()` (successor to the deepest-15%-band heuristic from #746) finds the *last* sample at or above a depth threshold and measures from there back to the first sample in the profile:

```dart
// lib/features/dive_log/domain/services/bottom_time_calculator.dart
int? ascentStart;
for (var i = sorted.length - 1; i >= 0; i--) {
    if (sorted[i].depth >= threshold) {
        ascentStart = sorted[i].timestamp;
        break;
    }
}
if (ascentStart == null) return null;

final bottomSeconds = ascentStart - sorted.first.timestamp;
return bottomSeconds > 0 ? bottomSeconds : null;
```

Nothing here bounds the result against the dive's own reported total duration. On a very shallow dive, the depth threshold sits only a little above the surface (it's `min(max(6, 33% of max depth), 85% of max depth)`, so for a 1.77 m dive that's ~1.5 m). If the profile carries samples for longer than the actual dive — idle/surface logging before or after it, which some drivers include in the same stream — a late sample crossing that low threshold from wave action or sensor noise becomes the new `ascentStart`, and bottom time comes out spanning however long that extra logging lasted.

Concretely: a 13-second dive computer duration produced a bottom time of 368 seconds — bottom time nearly half a minute longer than the entire dive, per its own reported runtime.

## Why it matters

Bottom time is descent-to-start-of-final-ascent by definition, which makes it a strict subset of total dive time — there's no legitimate scenario where it should exceed the dive's own duration. When it does, it's a visible, self-contradictory number on the dive detail page (a dive can show "0 min" runtime and "6 min" bottom time in the same header), and it also feeds anywhere else that consumes bottom time (trip/logbook stats, `DiveTanks` carry-over context, PDF exports).

## Fix

[...]

PR #1649, opened 2026-09-07
fix(safety-review): recompute the review when the profile changes und…

### LINKED ISSUE #1641: Re-parsing a dive leaves its stale safety review in place
# Re-parsing a dive leaves its stale safety review in place

`ReparseService.reparseDive()` rewrites a dive's profile series (new samples, corrected depths, whatever the fresh parse produces), but never touches its stored safety review. The next time the dive is viewed, `safetyReviewProvider`'s compute-through-cache serves the old review verbatim:

```dart
// lib/features/dive_log/presentation/providers/safety_review_providers.dart
final stored = await repo.getReview(diveId);
if (stored != null &&
    stored.engineVersion >= SafetyReviewService.engineVersion) {
  return stored;
}
```

This check only looks at `engineVersion`. It has no way to know the profile underneath it just changed, so it returns the stored review unconditionally whenever the version is current — regardless of whether a re-parse happened five seconds ago.

## Where the two paths diverge

`DiveRepository.editProfile` and a fresh download in `DiveComputerRepository` both already call `SafetyFindingsRepository.clearReviewForDive()` after rewriting a profile, specifically so the next view recomputes:

```dart
// lib/features/dive_log/data/repositories/dive_repository_impl.dart
// Profile changed: drop the stored safety review so it recomputes.
await SafetyFindingsRepository.clearReviewForDive(
  _db,
  _syncRepository,
  diveId,
);
```

`ReparseService.reparseDive()` (`lib/features/dive_computer/data/services/reparse_service.dart`) is the third path that rewrites a profile, and it's missing this call entirely.

## Why it matters

This is the same bug class as #1207 and #1208 (re-parse leaving `dive_data_sources` fields stale) — one more thing re-parse forgot to refresh. It's easy to hit in practice: a bad parse produces wrong findings (a depth-corrupted profile computing an impossible ppO2 or gradient-factor value, for instance), the user fixes the underlying data and re-parses expecting the findings to follow, and they don't. Nothing in the UI indicates the review is stale — it just looks like the fix didn't work.

It also compounds across a logbook: `SafetyReviewSweep` ("Analyze all dives") processes dives oldest-first because `profileAnalysisProvider`'s tissue/CNS/OTU chain carries forward between them, and that chain does invalidate correctly when an earlier dive's profile changes. But the sweep still has to actually recompute *each* dive's review to pick up a corrected upstream chain, and every dive with its own already-current-engine-version stored review hits this same stale-cache short-circuit — so re-parsing one dive early in a logbook doesn't fix findings on the dives after it, even when the underlying tissue-loading math is now correct.

## Fix

Have a fix ready: `reparseDive()` calls `SafetyFindingsRepository.clearReviewForDive()` once after the loop, if at least one source's re-parse succeeded — mirroring the other two paths. Comes with a test asserting a stale review (right engine version, wrong findings) gets replaced by a fresh one after `reparseDive()` runs, and a second test confirming a review is left untouched when every source in the reparse fails (nothing actually changed, so nothing should be invalidated). Happy to submit as a PR.

### LINKED ISSUE #1643: "Analyze all dives" doesn't force a genuine recompute for dives that already have a review
# "Analyze all dives" doesn't force a genuine recompute for dives that already have a review

`SafetyReviewSweep.run()` (the engine behind Settings → Safety Review → "Analyze all dives") invalidates and re-reads `safetyReviewProvider` for every dive:

```dart
// lib/features/dive_log/presentation/providers/safety_review_sweep.dart
_ref.invalidate(safetyReviewProvider(diveId));
await _ref.read(safetyReviewProvider(diveId).future);
```

`invalidate()` only clears Riverpod's cache of the provider's *last result*. The provider body itself still runs its own compute-through-cache check first:

```dart
// lib/features/dive_log/presentation/providers/safety_review_providers.dart
final stored = await repo.getReview(diveId);
if (stored != null &&
    stored.engineVersion >= SafetyReviewService.engineVersion) {
  return stored;
}
```

So for any dive that already has a stored review at the current engine version — which, after the first sweep, is every dive — this reads it straight back from the DB and returns it unchanged. The sweep runs to completion, reports every dive as swept, and nothing is actually recomputed.

## Why it matters

This is the tool a user reaches for specifically to force a bulk recompute — after fixing bad input data (a corrected re-parse, a manually edited profile) and expecting the whole logbook's findings to catch up. Right now it can't do that for anything that already has a review, which defeats its purpose. It also means a genuinely stale review — one left behind by a bug like #1641, or a raw import that predates a parser fix — cannot self-heal by running this action; only a bulk clear-then-recompute (which doesn't exist as a user-facing option today) or per-dive `SafetyFindingsRepository.clearReviewForDive()` calls would do it.

## Fix

Have a fix ready: the sweep drops the stored review (`SafetyFindingsRepository.clearReview()`, a new instance-method wrapper over the existing static `clearReviewForDive`) *before* invalidating and re-reading, for every dive it visits — guaranteeing `safetyReviewProvider`'s `stored == null` branch runs and a fresh computation actually happens, matching what "Analyze all dives" is supposed to do. Comes with a test that seeds a stale-but-current-engine-version review with wrong findings and asserts the sweep replaces it with the fixture's real findings, not just leaving it in place. Happy to submit as a PR — this pairs naturally with #1641 (`ReparseService` not invalidating a review at all).

urbamax:
## Summary

Two paths leave a dive's stored safety review in place after the data it was computed against changes, so it keeps showing findings from the old profile -- wrong ppO2, gradient-factor and missed-stop observations -- with nothing in the UI marking it stale.

Diagnosed by @trustthegoose (#1641, #1643). Both bite hardest right after a corrected re-parse: re-parse a dive whose profile had a bad depth spike, and the stale review still carries the impossible ppO2 / GF findings that spike produced.

Fixes #1641, fixes #1643.

## Changes

- **`ReparseService.reparseDive()`** rewrites the profile and the dive's computer-authored fields (gradient factors, ppO2, deco algorithm) but never touched the review. It now calls `SafetyFindingsRepository.clearReviewForDive()` once, after the loop, when at least one source's re-parse landed -- the same invalidation `editProfile` and the live-download path already do. A re-parse where every source fails changes nothing and leaves the review alone.
- **`SafetyReviewSweep`** ("Analyze all dives") only `invalidate`d the Riverpod provider. `safetyReviewProvider` still served each already-current-engine-version review straight from the DB, so after the first sweep the action recomputed nothing. It now drops each stored review first, via a new `SafetyFindingsRepository.clearReview()` instance wrapper over the existing static, so the provider's compute branch runs.

## Test Plan

- `reparse_service_test.dart`: a landed re-parse drops a seeded stale review (marker + findings); a re-parse where every source throws leaves it in place.
- `safety_review_sweep_test.dart`: with a repo that serves a current-engine-version review for every dive, the sweep still calls `clearReview` and re-`saveReview`s each one (before the fix it saved nothing).
- `flutter test test/features/dive_computer/ test/features/dive_log/` green; `flutter analyze` clean; `dart format` clean. (`deco_filter_providers_test` fails on `main` too, unrelated.)

### DIFF
--- a/lib/features/dive_computer/data/services/reparse_service.dart
+++ b/lib/features/dive_computer/data/services/reparse_service.dart
@@ -6,6 +6,7 @@ import 'package:submersion/core/data/repositories/sync_repository.dart';
 import 'package:submersion/core/database/database.dart';
 import 'package:submersion/features/dive_computer/data/services/libdc_dive_mode.dart';
 import 'package:submersion/features/dive_log/data/repositories/profile_series_repository.dart';
+import 'package:submersion/features/dive_log/data/repositories/safety_findings_repository.dart';
 import 'package:submersion/features/dive_log/data/repositories/tank_pressure_series_repository.dart';
 import 'package:submersion/features/dive_log/domain/codecs/profile_sample.dart'
     as codec;
@@ -369,6 +370,7 @@ class ReparseService {
     final sources = await getSourcesForDiveReparse(diveId);
     final errors = <String>[];
     var profilesPreserved = 0;
+    var appliedAny = false;
 
     for (final source in sources) {
       if (source.descriptorVendor == null ||
@@ -392,10 +394,25 @@ class ReparseService {
           descriptorModel: source.descriptorModel,
           libdivecomputerVersion: source.libdivecomputerVersion,
         );
+        appliedAny = true;
         if (outcome.profilePreserved) profilesPreserved++;
       } catch (e) {
         errors.add(e.toString());
       }
+    }
+
+    // A re-parse that landed rewrites the profile and/or the dive's
+    // computer-authored fields (gradient factors, ppO2, deco algorithm) the
+    // stored safety review was computed against, but never touched the review
+    // itself -- so it kept showing findings from the old data. Drop it here,
+    // once, the same as editProfile and the live-download path do (#1641).
+    // Left alone when every source failed: nothing changed, nothing to redo.
+    if (appliedAny) {
+      await SafetyFindingsRepository.clearReviewForDive(
+        db,
+        SyncRepository(database: db),
+        diveId,
+      );
     }
 
     return (errors: errors, profilesPreserved: profilesPreserved);
--- a/lib/features/dive_log/data/repositories/safety_findings_repository.dart
+++ b/lib/features/dive_log/data/repositories/safety_findings_repository.dart
@@ -263,6 +263,14 @@ class SafetyFindingsRepository {
     }
   }
 
+  /// Instance wrapper over [clearReviewForDive], for a caller that holds a
+  /// repository rather than a `(db, sync)` pair -- the "Analyze all dives"
+  /// sweep, which drops each stored review so the recompute actually runs
+  /// instead of the compute-through-cache handing back the current one
+  /// unchanged (#1643).
+  Future<void> clearReview(String diveId) =>
+      clearReviewForDive(_db, _syncRepository, diveId);
+
   /// Maps a stored finding row to its domain entity, or null when the row's
   /// rule_id does not correspond to a known [SafetyRuleId]. Callers drop null
   /// rows rather than surface a coerced (misleading) rule.
--- a/lib/features/dive_log/presentation/providers/safety_review_sweep.dart
+++ b/lib/features/dive_log/presentation/providers/safety_review_sweep.dart
@@ -93,7 +93,15 @@ class SafetyReviewSweep {
         );
       }
       try {
-        // Invalidate first. safetyReviewProvider is not autoDispose, so any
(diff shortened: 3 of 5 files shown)


urbamax:
Closing in favour of @trustthegoose's urbamax/submersion#2, which covers the same two bugs plus #1642, and — from validating on a physical device — also catches `ReparseService.reparseAllForComputer()` (the "Reparse All Dives" button), which calls `applyParsedUpdate()` directly and never goes through `reparseDive()`. That path had the same stale-review problem and this PR missed it.

His PR targets `urbamax:suunto-nautic`; these three are general reparse/safety-review bugs (not Suunto-specific), so they'll want forwarding to `main` too once that branch lands.

Issue #1682, opened 2026-09-09
Cell linearity line ignores the diver's decimal separator


ericgriffin:
The pre-dive cell linearity line renders its numbers with raw Dart string
interpolation and `toStringAsFixed`, both of which always emit '.' as the
decimal separator. A diver in a comma-decimal locale (de, es, fr, it, nl, pt,
hu) sees "Air 45.0 mV, expected 52.3 mV" where every number they type into the
app uses a comma.

In `lib/features/pre_dive/presentation/widgets/session_item_tile.dart`:

```dart
? l10n.preDive_runner_linearityLine(
    '$air',                        // raw interpolation
    expected.toStringAsFixed(1),   // ASCII '.', pinned to one decimal
    percent.round().toString(),    // int, no separator, fine as is
  )
```

The sibling value line on the same tile was fixed separately by routing it
through `formatDecimalForDisplay` in `lib/core/utils/number_input.dart`, a
locale-aware display formatter that preserves the value's own precision.

Fix:

- `air` goes through `formatDecimalForDisplay`.
- `expected` needs a fixed-digit display variant. `formatFixedForInput` in the
  same file already produces exactly the right output, but its name marks it as
  an input-seeding helper. Either add a `formatFixedForDisplay` twin and have
  `formatFixedForInput` delegate to it, or rename to make the shared intent
  clear.
- `percent` is an int and needs nothing.

Add a widget test pinned to a comma-decimal locale. Note that
`formatDecimalForDisplay` resolves `Intl.defaultLocale`, a mutable process
global that the app sets at `lib/app.dart:387` and that a widget test pumping
the tile alone never sets. Setting only the `MaterialApp` locale produces a
test that passes against unfixed code. Set both, with setUp/tearDown
save-restore, following
`test/features/dive_log/presentation/widgets/gas_input_locale_test.dart`.

Scope is display only. Nothing stored or parsed changes.

This is blocked on the cell linearity feature landing (branch
`ericgriffin/issue-986-brainstorm-245001`, issue #986).

PR #1686, opened 2026-09-09
fix(i18n): localise the decimal separator in unit display (#1683)

### LINKED ISSUE #1683: UnitFormatter emits ASCII decimal separators regardless of locale
`UnitFormatter` renders every number through `toStringAsFixed`, which
implements the Dart literal grammar and so always emits '.' as the decimal
separator. There are 21 such calls in `lib/core/utils/unit_formatter.dart`, and
exactly one use of `NumberFormat` (in `formatAltitudeWithGroup`, and only for
grouping).

The result is that a diver in a comma-decimal locale (de, es, fr, it, nl, pt,
hu) types "12,5" into a depth field and the app reads it back as "12.5 m".
Input was made locale-aware in #1091; display was not, so the two halves of the
app now disagree with each other.

Affected surfaces include depth, temperature, pressure, volume, tank volume,
SAC, RMV, weight, height, altitude, barometric and surface pressure, wind
speed, and speed. That is most numeric output in the app.

`lib/core/utils/number_input.dart` already has the pieces: a locale cache keyed
on `Intl.getCurrentLocale()`, a private separator swap, and, as of the pre-dive
value line fix, a `formatDecimalForDisplay` display formatter. The likely shape
of the fix is a fixed-digit display helper beside it, with `UnitFormatter`
routing its `toStringAsFixed` results through it.

Two things to watch:

- `NumberFormat` is the wrong tool for the digits. It renders the exact binary
  value, so raising its three-digit fraction cap far enough to stop it rounding
  makes it emit noise instead: 12.05 becomes "12.050000000000001". The existing
  helpers localise the separators of `toString`/`toStringAsFixed` output rather
  than formatting through `NumberFormat`, and a fix here should do the same.
- `UnitFormatter._trimTrailingZeros` matches a literal '.', so it has to run
  before localisation, not after. Under de the '.' it would match afterwards is
  the grouping separator.

Tests need to pin `Intl.defaultLocale`, not just a `MaterialApp` locale. The
formatter resolves the process global, which the app sets at
`lib/app.dart:387`; a test that sets only the widget locale will pass against
unfixed code.

Worth sizing before starting: this touches most numeric display in the app, and
a large number of existing tests assert ASCII output under an implicit locale.

ericgriffin:
Fixes #1683.

`UnitFormatter` rendered every number through `toStringAsFixed`, which implements the Dart literal grammar and so always emits `.`. Input was made locale-aware in #1091 but display was not, so the two halves of a numeric field disagreed with each other: a diver in a comma-decimal locale (de, es, fr, it, nl, pt, hu) typed "12,5" into a depth field and read back "12.5 m".

## What changed

- **`lib/core/utils/locale_number_symbols.dart` (new).** The locale-keyed `NumberFormat` cache and the separator swap, lifted out of `number_input.dart`. Both halves of the app's numeric text handling now resolve against one instance; two private caches would have been a slow way to reintroduce the #1091 mismatch.
- **`lib/core/utils/number_display.dart` (new).** `formatFixedForDisplay` and `localiseDecimalText`.
- **`lib/core/utils/unit_formatter.dart`.** All 21 fixed-digit renders route through the display formatter: depth, temperature, pressure, volume, tank volume, SAC, RMV, weight, barometric and surface pressure, wind speed, speed, and distance.

Display needed its own helper rather than reusing `formatDecimalForInput`. That one drops a trailing `.0` on purpose, because a text field wants "1250" rather than "1250.0" — but a rendered "200.0 bar" says the reading was taken to a tenth, so the zero is information. Routing display through the input helper has already silently changed rendered text once, so the two now live in separate files that cross-reference each other.

## Two ordering constraints

- `_trimTrailingZeros` matches a literal `.`, so it runs on the raw `toStringAsFixed` output and hands the result to the localiser afterwards. The other way round it does nothing under a comma locale, and under a grouping locale the `.` it would match is the group separator. A first pass at this change got the order wrong and rendered "26,0°C" where "26°C" was correct; the test for that specific hazard is what caught it.
- The digits still come from `toStringAsFixed`, not `NumberFormat`. NumberFormat renders the exact binary value, so a fraction cap raised far enough to stop it rounding starts emitting noise instead: 12.05 becomes "12.050000000000001".

## Deliberately unchanged

- **No grouping separators are introduced.** Grouping is a separate decision from the decimal separator and is not what the issue reports. `formatAltitude` keeps grouping through its own `NumberFormat` pattern.
- **Coordinates keep their dot-based text.** It is an internal source of truth shared with the DMS/DDM/UTM/MGRS parsers, and a comma is also the lat/lon separator there. There is a test asserting they stay ASCII under `de`.
- **CSV export is untouched.** It calls `toStringAsFixed` on the raw domain values directly, never through `UnitFormatter`, so a comma-locale export cannot become a broken comma-delimited file. PDF export does route through `UnitFormatter` and is now localised, which is correct for human-readable output.
- **Integer-only paths** (`formatHeight`'s whole centimetres, `formatGeoDistance`'s whole m/ft) are left alone: they have no decimal separator to get wrong.

## Testing

- `test/core/utils/unit_formatter_locale_test.dart` — 12 assertions that were red against the unfixed code, covering every affected surface plus the trim-ordering hazard, a dot-locale no-change guard, and the coordinate carve-out.
- `test/core/utils/number_display_test.dart` — the helpers themselves, including trailing-zero retention, absence of grouping, and the NumberFormat binary-noise case.

Both pin `Intl.defaultLocale`, the process global the app sets from the diver's language at `lib/app.dart`, rather than a `MaterialApp` locale. A widget-locale test passes against the unfixed code, so it would prove nothing here.

Because the new tests first went red only as a compile error, the helpers were mutation-tested afterwards: stubbing the localiser to `return text` fails 19 of the 25 assertions and leaves exactly the 6 that should survive (the dot-locale and coordinate guards).

`flutter analyze` is clean across the project, and the full suite is green: 25,702 passed, 21 skipped, 0 failures.

## Not in scope

There are roughly 459 other `toStringAsFixed` calls in `lib/` outside `UnitFormatter`. Many are calculations rather than display, but a real subset renders numbers directly (dive_log spans 27 files, dive_sites 12, statistics 8). The issue scopes to `UnitFormatter` and so does this PR; numeric display is not fully locale-correct until those are triaged separately.


ericgriffin:
Heads up for whoever merges these two: this PR and #1687 collide.

#1687 (issue #1682, cell linearity display) is stacked on #1679 and targets
`ericgriffin/issue-986-brainstorm-245001`, so it was cut before this branch
existed and neither `number_display.dart` nor `locale_number_symbols.dart` was
visible to it. It independently added its own display helpers to
`lib/core/utils/number_input.dart`:

- `formatFixedForDisplay(double, int)`, same name and signature as the one
  this PR puts in `number_display.dart`. Duplicated once both land, and an
  ambiguous import for any file that imports both.
- `formatDecimalForDisplay(double)`, which has no equivalent here. It keeps the
  value's own precision including a trailing zero, so `48.0 mV` stays
  `48,0 mV` rather than collapsing to `48 mV` the way `formatDecimalForInput`
  would.

Both PRs also rewrite `number_input.dart`, this one by moving the locale cache
and separator swap out to `locale_number_symbols.dart`, so a textual conflict
there is certain.

Resolution when the second one merges is a merge rather than a pick-one: keep
`formatDecimalForDisplay`, drop one copy of `formatFixedForDisplay`, and settle
on `number_display.dart` as the home for both.

ericgriffin:
Correcting my earlier comment here: the overlap with #1687 is smaller than I
described.

#1684 has since landed `formatDecimalForDisplay` on `main`, so #1687 no longer
defines its own and has been slimmed. The remaining overlap between this PR and
#1687 is just `formatFixedForDisplay(double, int)`, which both add with the same
signature, this one in a new `lib/core/utils/number_display.dart` and #1687 in
`number_input.dart`. Whichever merges second should drop its copy rather than
leave the symbol defined twice, since a file importing both would get an
ambiguous import. `number_display.dart` is the more natural home.

The conflict in `number_input.dart` still stands, as this PR moves the locale
cache and separator swap out to `locale_number_symbols.dart`.

PR #1739, opened 2026-09-10
feat(import-wizard): add shift-click multi-select

urbamax:
Adds shift-click multi-select to the dive import view, allowing users to rapidly check or uncheck a continuous block of dives (or other entities) without clicking every row. 

## What's here

| Piece | Notes |
|---|---|
| `ImportWizardNotifier.setSelections()` | A new method to perform a bulk add/remove of a `Set<int>` of indices in a single state update, respecting duplicate/skip boundaries. |
| `EntityReviewList` | Added `_lastToggledIndex` as a persistent variable to anchor the range. Reads `HardwareKeyboard.instance.logicalKeysPressed` directly in `onToggle` to detect Shift. |
| Range calculation | `EntityReviewList` knows the sorted `visualOrder` (via `nonDuplicateIndices`). When Shift is held, it finds the visual slice between the last clicked index and the current one, and passes it to `onSetSelections`. |

## Testing

`dart analyze` + `dart format` are clean.


urbamax:
@ericgriffin CI is fully green here (all checks passing, no conflicts) if you get a chance to take a look.

PR #1773, opened 2026-09-11
fix(sync): stage the parent dive for registry backfills and dive re-parses

ericgriffin:
## Summary

`dive_tanks`, `dive_data_sources`, `dive_profile_events` and `gas_switches` have no clock of their own. `SyncDataSerializer` exports them only for dives whose `hlc` is newer than the watermark, and none of them is in `SyncRepository.hlcTargets`, so `markRecordPending('diveTanks', ...)` stamps nothing. Two writers edited these rows without staging the parent dive, so their changes never reached a peer in an incremental changeset.

1. **Transmitter registry backfill** (`TransmitterRepository.applyToExistingDives`) filled volume, working pressure, material, preset, gear link, name and role onto matching tanks and staged only the tanks. It now stages each distinct touched dive once, inside the same transaction. This follows the same approach as #1769 did for `EquipmentRepository.deleteEquipment`.
2. **Dive re-parse** (`ReparseService.applyParsedUpdate`) rewrote the dive row, its source row, tanks, events and gas switches and staged nothing at all. It also deleted events, gas switches and dropped tanks without tombstones. A peer's import only upserts those rows, so staging the dive alone would have left every replaced event and switch on the peer beside its re-inserted copy. The fix therefore does both.

## Changes

- `TransmitterRepository.applyToExistingDives`: `markRecordPending('dives', ...)` once per touched dive after the per-tank marks.
- `ReparseService.applyParsedUpdate`:
  - stages the dive at the end of the transaction, unconditionally. A non-primary re-parse still rewrites its `dive_data_sources` row, which rides the same clock.
  - tombstones every event and gas switch it clears, and every tank it drops.
  - calls `SyncEventBus.notifyLocalChange()` after commit. A non-primary re-parse writes no series, so nothing announced it before.
- New `reparse_service_sync_test.dart`, kept separate because `reparse_service_test.dart` is already over 4,000 lines.

I audited every other `dive_tanks` writer (dive repository single and bulk paths, split, uncombine, merge, consolidation, tank pressure source exchange, dive computer import and merge). Each one already stages the parent dive.

## Test Plan

- [x] New tests record the watermark from a base export, run the write, then check the incremental changeset. For the backfill: only the touched dive is staged, an untouched dive's HLC does not move, and the filled tank rows are exported. For the re-parse: the dive, source row, tanks, events and gas switches are exported; tombstones for the removed tank, event and gas switch ride the changeset; a non-primary re-parse still advances the dive's HLC; the change is announced.
- [x] Each new test failed before its fix. Mutation checks (dropping the event tombstones, dropping the gas switch tombstones, or gating the dive mark on `isPrimary`) each turn one test red.
- [x] `flutter test` on the dive computer, transmitter, sync, consolidation, merge and re-parse menu suites: 1,481 passed
- [ ] Manual testing on: not run; the change is data-layer only

### DIFF
--- a/lib/features/dive_computer/data/services/reparse_service.dart
+++ b/lib/features/dive_computer/data/services/reparse_service.dart
@@ -4,6 +4,7 @@ import 'package:uuid/uuid.dart';
 
 import 'package:submersion/core/data/repositories/sync_repository.dart';
 import 'package:submersion/core/database/database.dart';
+import 'package:submersion/core/services/sync/sync_event_bus.dart';
 import 'package:submersion/features/dive_computer/data/services/libdc_dive_mode.dart';
 import 'package:submersion/features/dive_log/data/repositories/profile_series_repository.dart';
 import 'package:submersion/features/dive_log/data/repositories/tank_pressure_series_repository.dart';
@@ -46,6 +47,7 @@ class ReparseService {
     ProfileSeriesRepository? profileSeries,
     TankPressureSeriesRepository? tankSeries,
   }) : _transmitterMatcherLoader = transmitterMatcherLoader,
+       _sync = SyncRepository(database: db),
        _profileSeries =
            profileSeries ??
            ProfileSeriesRepository(
@@ -59,6 +61,7 @@ class ReparseService {
              syncRepository: SyncRepository(database: db),
            );
 
+  final SyncRepository _sync;
   final ProfileSeriesRepository _profileSeries;
   final TankPressureSeriesRepository _tankSeries;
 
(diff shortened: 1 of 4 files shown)


ericgriffin:
Closing without merging. Staging a clockless child's parent dive stamps a fresh HLC on the whole dive row, and dives merge whole-row by HLC, so a device holding a stale snapshot could overwrite a peer's newer dive edits. #1769 instead makes the incremental export honour pending records for clockless children, which covers the transmitter backfill's existing `diveTanks` marks.

The re-parse gap is still open under that design: `ReparseService.applyParsedUpdate` marks none of the child rows it writes pending, and deletes events, gas switches and dropped tanks without tombstones. Commit 96a66022 on `ericgriffin/fix-dive-tank-sync-staging` has the tombstone half and tests in `reparse_service_sync_test.dart`, left on the branch for reuse.

Issue #755, opened 2026-07-27
Remove the Computer option from the ceiling line; always show the computed ceiling


readme42:
## Summary

The ceiling line has a **Computer vs Calculated** source toggle, but the **Computer** option is redundant and adds no information. No dive computer exports a continuous ceiling — every import path only provides the stepped **stop depth**, which is stored in `dive_profiles.ceiling`. So "Computer ceiling" renders the same stepped data as the **deco-stop band (Computer)**. The only thing the ceiling line uniquely contributes is the **exact, continuous curve**, which is inherently a Calculated value.

Proposal: **drop the Computer option from the ceiling line and always render the computed (Calculated) ceiling.** Keep the Computer/Calculated toggle on the **deco stops**, where the stepped schedule genuinely differs between the dive computer and the app's recomputation.

## Why (verified)

Across every source, the stored `ceiling` is the computer's stepped stop depth, never a continuous ceiling:

- Direct DC download (libdivecomputer): `deco_depth` -> `ceiling` (`parsed_dive_mapper.dart` ~L70, `s.decoType != 0 ? s.decoDepth : null`). The native sample has no exact-ceiling field.
- Subsurface XML: `stopdepth` -> `ceiling` (#677).
- UDDF: `decodepth` -> `ceiling` (#750).
- Shearwater: `decoDepth` -> `ceiling`.

The deco-stop band is that same ceiling curve quantized to stops (`deco_stop_curve.dart`). On the **Computer** source the ceiling is already stepped, so the ceiling line and the deco-stop band are pixel-identical — redundant.

## Distinct data (the 4 toggle cells collapse to 3)

| | Ceiling line | Deco-stop band |
|---|---|---|
| Computer | stepped stops | stepped stops (identical) |
| Calculated | smooth exact curve | stepped (quantized) |

"Ceiling / Computer" is the duplicate cell.

## Suggested change

- **Ceiling line:** remove the Computer option; always compute/show the exact ceiling.
- **Deco stops:** keep the Computer vs Calculated toggle (the meaningful comparison: DC's actual stops vs app-recomputed stops).

Touch points: `ceilingSource` in `profile_legend_provider.dart`; the ceiling source control in `dive_profile_legend.dart` (~L651); the `ceilingSource == computer && hasComputerCeiling` gating in `profile_analysis_provider.dart` (~L413-421).

### Scope options

[...]

PR #1794, opened 2026-09-12
refactor(dive-log): always calculate the profile's deco metrics (#1793)

### LINKED ISSUE #767: Retire defaultCeilingSource setting + column (schema migration follow-up to #755/#761)
## Background

#761 removed the Computer/Calculated source toggle from the ceiling line (the ceiling line now always renders the app-computed continuous curve, since no dive computer or import format ever provides a continuous ceiling — only stepped stop depths, which the deco-stop band already shows). See #755 for the full rationale.

That PR intentionally left the persistence layer in place to avoid a schema migration. `defaultCeilingSource` is now **dead storage**: written and synced, but read by nothing that affects behavior.

## Scope

Retire the setting end-to-end. Bundle this into the **next `diver_settings` schema migration done for another reason** — do not do a standalone schema-version bump just for this one inert column.

- [ ] Drop the `defaultCeilingSource` column from `diver_settings` (`lib/core/database/database.dart:1516`) as part of a batched migration
- [ ] Remove `setDefaultCeilingSource` notifier method (`lib/features/settings/presentation/providers/settings_providers.dart:1264`)
- [ ] Remove the `defaultCeilingSource` field from the settings model (`settings_providers.dart:188`, constructor, `copyWith`)
- [ ] Remove the read/write in `diver_settings_repository.dart` (lines ~105, ~255, ~446)
- [ ] Update / remove the sync seed entry (`sync_data_serializer.dart:5135`)
- [ ] Update tests that reference `defaultCeilingSource`

## Back-compat requirement

`defaultCeilingSource` is a **synced** field. After dropping the local column, the sync deserializer must **stay tolerant of the legacy key** so cross-version sync with older clients doesn't break. Ignore-on-read rather than reject.

## Migration checklist reminder

Follow the schema-bump discipline: guarded migration helper + backstop + tripwire test + sync-seed update. Verify all four are covered before merging.

### LINKED ISSUE #1793: Dive data source preference is confusing and its numbers are not trustworthy
**Describe the bug**

Settings > Decompression has a "Data Source Preferences" section offering
Calculated or Dive Computer for NDL, Deco Stop, TTS, GTR and CNS, and the
chart options dialog repeats the same choice as DC/Calc buttons on each of
those rows. The two sets of controls are confusing on their own, and the
numbers they produce are not trustworthy.

Picking Dive Computer shows whatever the computer happened to record, which
means the same dive reads differently depending on which computer was worn,
and a computer swap partway through a logbook makes the numbers incomparable
from one dive to the next. Some computers make it worse: a Cressi Leonardo
logs its deco obligation as a single bit and no numbers, so read literally it
claims the ceiling is at the surface and there is no no-stop time left.

The preference also does not do what the labels suggest. The ceiling line
ignores it entirely and always draws the calculated curve, and choosing Dive
Computer for a metric the computer never recorded silently falls back to
Calculated with nothing on screen saying so.

**To Reproduce**
1. Go to Settings > Decompression > Data Source Preferences
2. Set NDL Source and Deco Stop Source to Dive Computer
3. Open a dive from a computer that does not report those values
4. The profile shows calculated values with no indication the preference was
   not honoured
5. Open the same dive from a different computer and the numbers change

**Expected behavior**

One trustworthy set of numbers. Every decompression metric on the profile
should come from one model using the diver's own gradient factors, so the
values are comparable across a whole logbook and across a change of computer,
with no per-metric choice to get wrong.

**Platform details:**
- OS: all
- Version: v1.7.10

**Additional context**

Diagnosis. The preference is stored per metric as a `MetricDataSource`
(`default_ndl_source`, `default_ceiling_source`, `default_tts_source`,
`default_cns_source`, `default_deco_stop_source`, `default_gtr_source` on
`diver_settings`), seeded into per-dive legend state and resolved in
`overlayComputerDecoData`. The resolution carries a pile of special cases that
exist only to decide when computer data can be trusted: an all-zero series
counts as unreported (the Leonardo rule, PR #342), the deco stop band resolves
separately from the ceiling so one does not drag the other, and each metric
falls back independently. `MetricSourceInfo` was built to report which source
actually won, but nothing in the app ever read it, so the fallback is
invisible to the diver.

`default_ceiling_source` has been dead storage since #755/#761 removed the
ceiling toggle, and #767 asks for its column to be dropped in the next
`diver_settings` migration done for another reason. This is that migration,
so #767 is closed by the same change. `use_dive_computer_cns_data` is the v42
ancestor of `default_cns_source` and has been read by nothing since.

Settling on the calculated value removes the choice and the machinery behind
it rather than just hiding the UI.

dotanalon:
## Summary

Retires the per-metric Computer/Calculated data source preference. Every
decompression metric on the dive profile -- NDL, TTS, CNS, GTR and the deco
stop band -- is now the app's own Buhlmann calculation, so one model and the
diver's own gradient factors produce every number. The ceiling line already
worked this way since #755/#761.

The preference was two controls for one choice (a settings section and DC/Calc
buttons on each chart row), and the computer side of it needed a growing pile
of special cases to decide when a computer's readings could be trusted at all:
an all-zero series had to count as unreported to survive a Cressi Leonardo,
which logs its deco obligation as a single bit and no numbers. Choosing
Calculated lets all of that go rather than keeping it behind a hidden toggle.
`MetricSourceInfo` existed to report which source actually won, but nothing
ever read it, so the silent fallback was invisible anyway.

This also carries out #767: `default_ceiling_source` has been dead storage
since the ceiling toggle was removed, and that issue asked for its column drop
to ride along with the next `diver_settings` migration done for another
reason. This is that migration, so the other five source columns and the
vestigial `use_dive_computer_cns_data` (the v42 ancestor of
`default_cns_source`) go with it.

Note this changes what existing dives display: wherever a computer recorded
its own NDL, TTS, CNS, GTR or stop depths and that value was being shown, the
profile now shows the calculated value instead.

## Changes

- Remove `MetricDataSource`, `MetricSourceInfo` and `metricSourceInfoProvider`
- Remove the six `AppSettings` source fields, their setters and the
  "Data Source Preferences" settings section. The GTR reserve-pressure tile
  lived in that section and moves to Ascent, where it belongs
- Remove the five source fields and setters from `ProfileLegendState`, and the
  per-dive DC/Calc segmented controls from the chart options dialog, along
  with `buildToggleWithSource` and `sourceSegments`
- Collapse `overlayComputerDecoData` to `overlayRebreatherSensorData`: the
  computer overlay, the zero-sentinel probes and the Leonardo rule are gone;
  what remains is ppO2 / O2 sensor / O2 cell resolution, which is measured
  data rather than a model output and was never part of this feature
- Delete `computer_cns_extractor.dart`, now unreachable: the analysis no
  longer overrides `o2Exposure.cnsStart`/`cnsEnd` from the computer, and
  residual CNS always comes from the previous dive's calculated analysis
- Schema v208 drops `default_ndl_source`, `default_ceiling_source`,
  `default_tts_source`, `default_cns_source`, `default_deco_stop_source`,
  `default_gtr_source` and `use_dive_computer_cns_data`, and the deco-stop and
  GTR `beforeOpen` backstops stop re-asserting their source columns -- they
  run unconditionally on every open and would otherwise re-add on each launch
  what the rung just dropped. A tripwire test guards that
- Drop the retired sync seeds. Inbound payloads from older peers still carry
  the keys and are ignored on read rather than rejected, with a test pinning it
- Remove 11 now-unused l10n keys across all 11 locales

## Test Plan

- [ ] `flutter test` passes -- not run in full; see below
- [x] `dart format .` clean
- [x] `test/core` (database, sync, export/backup, utils): 5628 passed
- [x] `test/features/dive_log`: 4040 passed
- [x] `test/features/settings`: passed
- [ ] Manual testing on: none

The affected-tests selector resolved to 1922 files -- effectively the whole
suite, because the change touches `database.dart`, settings and l10n, which
nearly every test imports. Rather than run that, the three directories
covering the blast radius were run in full. Not covered: the remaining feature
directories, which reach the changed files only transitively.

`test/features/dive_log/presentation/providers/deco_filter_providers_test.dart`
fails on this machine both with and without these changes. It is timezone
dependent, not a regression: a fixture dive at `DateTime(2026, 1, 3)` local
stores as `2026-01-02T21:00Z` under UTC+3 and falls outside a `startDate` of
the same day. It passes under `TZ=UTC`.

No manual testing was done. Worth eyeballing a real deco dive from a computer
that reports its own NDL and stop depths before merging, since that is exactly
the profile whose displayed numbers change.

Fixes #1793
Fixes #767


ericgriffin:
Why would we not want to expose the dive computer data?

PR #1802, opened 2026-09-12
fix(list): wrap long dive and site card titles instead of truncating

ericgriffin:
## Summary

On a narrow window, the detailed Dives and Sites cards cut a long title off with an ellipsis, while the Trips and Equipment cards wrapped it onto more lines. This PR makes the Dives and Sites cards wrap too, so the full title is always visible and all four lists behave the same way.

## Changes

- `DiveListTile` (detailed dive card): removed `overflow: TextOverflow.ellipsis` from the title. The title had no `maxLines`, but `TextPainter` still ellipsizes the first overflowing line when `maxLines` is null, so that setting alone kept the title to one line. The forced strut it already uses keeps wrapped lines at the normal titleMedium line height.
- `SiteListTile` (detailed site card): removed `maxLines: 1` and the ellipsis from the title.
- The trailing title-row icons (favorite, rating, shared badge, and so on) stay vertically centered beside the title, the same as the trip card's shared badge.
- Compact and dense list modes are unchanged. All four features truncate the same way in those modes.
- Map-background card variants grow with their content (the map is a `Positioned.fill` behind it), so a taller title cannot overflow.

## Test Plan

- [x] New widget tests: at a 353px phone width, a long dive title and a long site title each lay out on more than one line with `didExceedMaxLines == false`. Both failed before the fix.
- [x] Existing Dive and Site card, list content, and dive_log page tests pass (573 tests)
- [ ] Manual testing on: checked with a throwaway golden render at 380px width (before: both titles ellipsized; after: both wrap onto three lines)

### DIFF
--- a/lib/features/dive_log/presentation/pages/dive_list_page.dart
+++ b/lib/features/dive_log/presentation/pages/dive_list_page.dart
@@ -967,13 +967,15 @@ class DiveListTile extends ConsumerWidget {
                                   // value (so the date line below doesn't
                                   // shift up) while textHeightBehavior only
                                   // repositions the ink within that space.
+                                  // No ellipsis: a long title wraps, like the
+                                  // trip and equipment cards, so the full
+                                  // name is always visible.
                                   style: titleStyle
                                       ?.copyWith(
                                         fontWeight: FontWeight.w600,
                                         color: primaryTextColor,
                                       )
                                       .inkCentered,
-                                  overflow: TextOverflow.ellipsis,
                                   textHeightBehavior:
                                       inkCenteredTextHeightBehavior,
                                   strutStyle: titleStyle?.preservingStrut,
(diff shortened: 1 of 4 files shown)


ericgriffin:
Re the suppressed Copilot note on `dive_list_page.dart` (dense dive mode now wraps the title): leaving this as is, no change.

Dives have no dense card anymore. `DenseDiveListTile` was removed on purpose in f1ea76fb0b9, and `dive_list_item.dart` routes `ListViewMode.dense` to the detailed `DiveListTile` as a fallback. Neither the Dives list menu nor the Appearance settings dropdown offers dense for dives, so the only way to reach it is a value saved before that removal. That path already renders the full detailed card, so wrapping its title is the intended behavior for the card actually shown. Adding a mode flag would make the same card look different depending on a setting the diver can no longer see. The one-line dense rows for Sites, Trips, and Equipment are separate widgets and are unchanged.

Issue #1855, opened 2026-09-13
A locked temp file during restore leaves the database closed until the app restarts, masking the real error


urbamax:
## Summary

A restore (from a local file or from a cloud backup, both go through the same `DatabaseService.restore` swap) can fail with something like:

```
PathAccessException: Cannot delete file '...submersion.db.pre-restore', ...
  (OS Error: Operation not permitted, errno = 1)
```

and afterward the app is left with **no open database** until it is restarted -- the restore's own error-recovery path never gets a chance to run.

## Root cause

In `DatabaseService.restore` (`lib/core/services/database_service.dart`):

1. `await _deleteIfExists(asidePath);` (cleanup of a stale `.pre-restore` file from an earlier failed restore) runs **after** `await close(strict: true)` but **outside** the try/catch that the swap below is wrapped in. If this delete throws -- e.g. the database lives in a location a cloud-sync provider (iCloud Drive, OneDrive, Dropbox, ...) is actively materializing/evicting files in, which briefly denies delete/rename with `EPERM`/errno 1 -- the exception propagates straight out of `restore()`. `initialize()` is never called again, so the live database stays closed.
2. Even when a stale file's delete happens to occur *inside* the guarded try (the analogous `stagingPath` cleanup in the swap-failure `catch` block, and the three cleanup deletes in the `DatabaseVersionMismatchException` handler), those calls use `_deleteIfExists` rather than the already-existing `_bestEffortDelete` helper. A transient failure there also skips the `initialize()` call right after it, for the same reason.

Both are cleanup of files the recovery path no longer needs (an orphaned staging copy, a too-new file about to be overwritten by the rollback rename anyway) -- their success was never required for the rollback to be correct, only convenient.

## Impact

- The restore fails with a confusing "cannot delete file" error instead of whatever caused it.
- Worse: the app is left with a closed database (calls into it start failing with "database not initialized" or similar) until the user force-quits and reopens the app -- there is no way to tell from the error that this happened.
- This reproduces reliably whenever the database (or the default `Documents/Submersion` folder) sits inside a folder managed by a cloud-sync file provider on the OS (confirmed with iCloud Drive's "Desktop & Documents" sync on macOS, which is a common default), since that is exactly the class of transient lock these delete calls can hit.

## Expected

A transient failure to remove a leftover temp file during restore recovery should not prevent the recovery from completing (rolling the original database back into place and reopening it) or mask the underlying error that triggered the recovery in the first place.

PR #1960, opened 2026-09-14
fix(backup): make automatic pruning resilient to an undeletable old backup

### LINKED ISSUE #1959: Restore aborts when the pre-restore safety backup's pruning hits an undeletable old backup
## Problem

`restoreFromFile` takes a safety backup of the live database before swapping the restore source in (`performBackup`), which then prunes old manual backups past the retention count (`pruneOldBackups`). The prune loop calls `deleteBackup` with no `try`/`catch`, so a single old backup this platform refuses to delete — a custom local folder outside the sandbox, a locked network share, whatever the cause — throws all the way up through `performBackup` and `restoreFromFile`, aborting the restore entirely before it ever reaches the file the diver actually picked.

Symptom: `Échec de la restauration : PathAccessException: Cannot delete file, path = '<some old backup file>' ...`, on a restore attempt that never touches the picked file at all.

Worse: a local-delete failure inside `deleteBackup` also skips `_preferences.removeRecord`, so the record stays in history and the same file gets retried — and fails — on every future backup or restore attempt, making this a permanent, repeating failure rather than a one-off.

This is a plausible second contributor to the original #1855 symptom, on top of the `.pre-restore` handling #1856 and #1916 already hardened — those fixed the swap step itself, but this fires earlier, during the safety-backup step that runs before the swap is ever reached.

## Repro

1. Configure a local backup location the app cannot reliably delete from (e.g. a custom folder where a file has since become locked/permission-denied for any reason).
2. Accumulate enough manual backups to exceed the retention count.
3. Attempt any restore. `performBackup`'s pre-restore snapshot triggers `pruneOldBackups`, which throws on the undeletable file and aborts the restore before it reaches the swap step.

urbamax:
Closes #1959

## The bug

`restoreFromFile` takes a safety backup of the live database before swapping the restore source in (`performBackup`), which then prunes old manual backups past the retention count (`pruneOldBackups`). The prune loop called `deleteBackup` with no `try`/`catch`, so a single old backup this platform refuses to delete — a custom local folder outside the sandbox, a locked network share — threw all the way up through `performBackup` and `restoreFromFile`, **aborting the restore entirely before it ever reached the file the diver actually picked**.

Worse, a local-delete failure inside `deleteBackup` also skips `_preferences.removeRecord`, so the record stays in history and the same file gets retried — and fails — on every future backup or restore attempt. This turns a one-off permission hiccup into a permanent, repeating failure.

I found this while re-verifying #1855 after #1856/#1916 landed: those hardened the swap step itself (the `.pre-restore` handling), but this fires *earlier*, during the safety-backup step that runs before the swap is ever reached — so a diver could still see the exact same class of error even with both of those fixes in place.

## The fix

Wrap the per-record `deleteBackup` call in `pruneOldBackups`'s loop in a `try`/`catch`. Pruning is a side effect of the operation that triggered it (a fresh backup, or the pre-restore safety snapshot), not the operation itself — one record that can't be deleted now logs a warning and is left for the next prune, instead of blocking the backup or restore currently in progress. `deleteBackup`'s own semantics are untouched, so an explicit user-initiated delete (the trash icon in backup history) still surfaces a real failure — only the automatic background pruning became best-effort.

## Tests

- New regression test: a read-only containing directory (matching the existing chmod-based pattern in `pre_migration_backup_service_test.dart`) makes one record's local delete throw `PathAccessException`; prune must still run to completion, remove the other eligible records, and leave the undeletable one in history for a retry. Mutation-checked against the unfixed code (fails without the fix, passes with it).
- Full `test/features/backup/` suite (450 tests) passes.
- `flutter analyze` clean project-wide.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ericgriffin:
Addressed the suppressed review note on the permission-based regression test in 47bed4e.

After the test sets the temp directory to mode 555, it now probes whether the directory still accepts a new entry (the same directory-write permission unlink needs). If it does, the run is root or chmod was unavailable or ignored, so the test calls markTestSkipped instead of asserting on a locked file that would in fact delete cleanly. This matches the probe in bundled_photo_exporter_test.dart.

Verified locally: as a normal user the test runs and passes; with the directory left writable the test reports as skipped rather than failing.

ericgriffin (approved):

PR #1967, opened 2026-09-14
Pre-dive checklist: show full service-clock values and link checks to dives by chronological order

### LINKED ISSUE #1895: Pre-dive checklist: show full service-clock values and link checks to dives by chronological order
**Summary**

Two related gaps in the pre-dive checklist feature:

1. When running a pre-dive checklist, a linked equipment item's maintenance intervals (e.g. a rebreather with a scrubber and CO2 sensors) only surface a short text line when a service is already **overdue**. The individual configured services and their concrete values (e.g. "Scrubber: 8 of 12 hours left", "CO2 sensor due 2026-09-20") are not shown at all.
2. Auto-linking a completed pre-dive check to the dive it belongs to is bound to a fixed 3-hour time window and a strict 1:1 relationship, instead of simply following chronological order, allowing several checks to link to the same dive, and re-splitting existing links when an older dive is imported afterwards.

**Version**

v1.7.10.8228 (analyzed against commit ef9b9814f on `main`)

**Steps to reproduce**

*Part 1 - service data in the checklist*
1. Create an equipment item (e.g. a rebreather) with several service schedules configured (scrubber: hour-based, CO2 sensors: date-based), each currently `ok` or `dueSoon`, none `overdue`.
2. Create a pre-dive checklist that includes this item (either directly or via an equipment set).
3. Start/run the checklist.

*Part 2 - chronological linking*
1. Perform two pre-dive checks with no dive imported between them.
2. Import a dive that occurred after both checks.
3. Afterwards, import an older dive that occurred chronologically between the two checks.

**Expected**

*Part 1:* The checklist row for the equipment item lists every configured service individually with its current concrete value - remaining amount per unit (hours, dives, cold-water cycles, etc.) and/or due date - regardless of whether its status is `ok`, `dueSoon`, or `overdue`. Once the item has been checked off, looking back at it still shows every service's value, frozen at the moment it was checked off (not only the ones that were overdue at that time).

*Part 2:* A completed check links, on import, to the chronologically next dive that follows it, with no time window. Several checks preceding the same dive all link to that one dive (n:1). When a dive is imported later that falls chronologically in the middle of an already-linked group of checks, the group is automatically re-split: each check ends up linked to its actual next dive, even if that means breaking apart a group link that already existed. A dive with no preceding check simply stays unlinked.

**Actual**

*Part 1:* Only a short trigger-text line is rendered, and only when at least one service is overdue. Checked-off items freeze the same overdue-only subset.

*Part 2:* `ChecklistDiveLinker` only links within `linkWindow = 3h` before and `forwardGrace = 15min` after the dive start, and its 1:1 guard (`getSessionForDive(diveId) != null`) prevents a second check from linking to a dive that already has one. Existing links are never re-evaluated once set, even when an older dive is imported afterwards.

**Root cause**

*Part 1:* `SessionItemTile._overdueEntries()` (`lib/features/pre_dive/presentation/widgets/session_item_tile.dart:70-89`) explicitly filters the live `serviceClockStatusesProvider(equipmentId)` results to `severity == ServiceClockSeverity.overdue` (line 84) and discards everything else; the same filtering carries into the frozen `OverdueServiceEntry` snapshot written when an item is checked off. The full per-service data already exists: `ServiceClockStatus.usageByUnit` (a `Map<ExposureUnit, ClockUsage>` with `interval`, `since`, `remaining`) and `dueDate`/`daysUntilDue` (`lib/features/equipment/domain/entities/service_clock_status.dart:71-166`). The equipment overview already renders all of this via `ServiceClocksCard` (`lib/features/equipment/presentation/widgets/service_clocks_card.dart`). The checklist item is already linked to its equipment through `PreDiveSessionItem.equipmentId` (set in `session_item_composer.dart:75,95`), including for equipment sets, where one session item is created per member device. No data linkage is missing - only the full values never reach the UI.

*Part 2:* `autoLinkForDive()` (`lib/features/pre_dive/data/services/checklist_dive_linker.dart:45-76`) is deliberately built around a time window and a 1:1 relationship (documented in the comments at lines 7-9 and 16-18), and only runs once per newly imported dive (`applyForImportedDive`), never re-evaluating links that were set on a previous import.

**Proposal**

*Part 1:* Extend `SessionItemTile` and the checked-off snapshot mechanism to carry and render the full `ServiceClockStatus` values (`usageByUnit`, `dueDate`) for every configured service of the linked equipment, not only the overdue subset. Reuse the existing formatting/rendering logic from `ServiceClocksCard` and `service_trigger_text.dart` where possible instead of duplicating it.

*Part 2:* Rework the matching logic in `ChecklistDiveLinker`: drop the `linkWindow`/`forwardGrace` time window and instead link each unlinked check to the chronologically next dive of the same diver; replace the 1:1 guard with n:1 support so multiple checks preceding the same dive all link to it. When a dive is imported, also re-evaluate existing links that fall within the affected time range and re-split any group whose members now have a closer next dive available, so every check ends up linked to its actual next dive.

**Open questions**

None - resolved during issue drafting:
- Checked-off items freeze all service values, not only overdue ones (Part 1).
- No cap on how long an unlinked check may wait for its next dive (Part 2).
- A newly imported dive that falls inside an existing multi-check group re-splits that group automatically (Part 2).

**Out of scope**

Changes to how service schedules themselves are calculated (`ServiceSchedule`, `ServiceClockStatus`).

alpheios-one:
## Related Issue

Closes #1895

## Summary

Two related fixes to the pre-dive checklist's equipment integration:

1. A checklist row for equipment with service clocks (scrubber hours, CO2 sensor date, etc.) now shows every configured clock with its concrete value, not only overdue ones.
2. Auto-linking a completed check to its dive now follows pure chronological order instead of a fixed time window and a 1:1 relationship.

## Changes

**Service-clock display**
- `OverdueServiceEntry` gains `severity` and `usageByUnit`, kept backward-compatible: a snapshot frozen before this change (missing both fields) decodes with `severity = overdue` and an empty `usageByUnit`, matching what it always meant. No DB migration — the column stays the same JSON-text field.
- `SessionItemTile` and the checked-off snapshot (`PreDiveSessionRunnerPage`) drop the `severity == overdue` filter and render every clock, colored by its own severity. The existing "Service overdue" heading is kept as-is and still only appears when at least one clock actually is overdue.
- Extracted the severity→color mapping out of `ServiceClocksCard` into a shared util (`service_clock_severity_display.dart`) so the checklist and the equipment detail page stay visually consistent.

**Chronological dive linking**
- Added `DiveRepository.getNextDive` (mirrors `getPreviousDive`'s predicate/ordering in reverse, with an explicit diver filter).
- Rewrote `ChecklistDiveLinker.autoLinkForDive`: a check now links to the diver's actual next dive, with no upper time bound. Several checks preceding the same dive all link to it (previously 1:1 only). Because every dive import re-evaluates every session of that diver against the new dive, importing an older dive afterwards correctly re-splits an existing group — no explicit "reconciliation" step needed.
- The forward-skew tolerance (a check completed slightly after the dive's recorded start, e.g. due to computer/phone clock drift) is raised from 15 minutes to 3 hours, to also cover a DST change between the check and the dive.
- All four import call sites (UDDF, dive-computer download, HealthKit, the import wizard provider) are untouched — `applyForImportedDive`/`autoLinkForDive`'s signatures didn't change.

## Test Plan

- [x] `flutter test` passes (full `pre_dive` and `equipment` suites, plus the new `dive_repository_next_dive_test.dart` and the rewritten `checklist_dive_linker_test.dart`)
- [x] Manual testing on: Windows desktop


alpheios-one:
@ericgriffin This PR is ready for review. Happy to adjust anything if something's open - just let me know.

Issue #1900, opened 2026-09-13
Numeric input fields silently discard invalid input across the app


alpheios-one:
## Summary

Numeric text fields across the app silently discard or reset invalid input instead of telling the diver what went wrong. This surfaced while working on the Trimix Mixer improvements (#1876): the mixer's price field turned out to have the same root cause as several other fields, and a closer look showed it is an app-wide pattern rather than a mixer-specific bug.

## Background

`parseUserDecimal`/`parseUserInt` (`lib/core/utils/number_input.dart`) deliberately return `null` for text that cannot be read as a number in the active locale - for example `"14.8"` under a comma-decimal locale (German, Spanish, Italian), where `.` is the grouping separator, not the decimal one. This is intentional and correct: it is the fix for #1091, where a lenient parser used to silently misread such input as a completely different number (`"6,4"` read as `64`).

The problem is what callers do with that `null`. Across the app, the overwhelming majority of call sites swallow it silently:

```dart
double pressureOrZero(String text) => parseUserDecimal(text) ?? 0;
```

```dart
double mixPercentOrKeep(String text, double previous) {
  if (text.trim().isEmpty) return previous;
  return parseUserDecimal(text) ?? previous;
}
```

A diver who types a `.` where the app expects a `,` gets no indication anything went wrong - the field just quietly resets to `0`, keeps its previous value, or (in a couple of spots) stores `null` directly into a repository.

## Scope

A search for `parseUserDecimal`/`parseUserInt` turns up roughly 190+ call sites across 53 files, spanning nearly every feature: dive log editing, dive sites, equipment, the dive planner, tank/weight presets, transmitters, CCR/SCR settings, and more. A representative sample:

- `lib/features/dive_log/presentation/pages/dive_edit_page.dart` (multiple lines around 1189-1211, 5041-5152) - depth, water temperature, wind, humidity, weighting, all falling back with `?? 0`.
- `lib/features/dive_log/presentation/widgets/tank_editor.dart` (183-331, 723-737) - fill pressure, volume, O2/He, falling back with `?? 21.0`/`?? 0.0`.
- `lib/features/dive_sites/presentation/pages/site_edit_page.dart` (735, 1675-1677) - min/max depth, altitude.
- `lib/features/equipment/presentation/pages/equipment_edit_page.dart:979` - purchase price stores `null` directly, no feedback.
[...]

PR #1972, opened 2026-09-14
fix(dive-log): auto-assign 100% O2 when a cylinder's role is set to oxygen supply

### LINKED ISSUE #726: Dives / Dive / Gas & equipment / Cylinder – Automatically assign oxygen
Android + Windows 1.6.0.114

If I add a cylinder during the dive, I can select ‘Role’ -> ‘Oxygen’ but then have to enter ‘100% oxygen’ separately.
Having to enter this manually after selecting ‘Oxygen’ as the cylinder’s role makes no sense.
The correct approach would be for the oxygen percentage to be set automatically to 100% when you select ‘Role’ -> ‘Oxygen’.
 
Image

alpheios-one:
## Related Issue

Closes #726

## Summary

A CCR oxygen supply cylinder feeds pure O2 into the loop by definition, so
having to type "100" by hand after picking `Role -> Oxygen Supply` added
nothing.

## Changes

- **Manual dive editor** (`tank_editor.dart`): selecting `Oxygen Supply` in
  the tank's role dropdown now pre-fills 100% O2 / 0% He, the same pattern
  already used for gas templates and tank presets. The fields stay editable
  afterward.
- **Dive-computer import** (`parsed_tank_resolver.dart`): a related, more
  serious bug found while investigating this. An HP CCR oxygen tank has no
  gas mix of its own (Shearwater never links one; it only ever tags a
  gasmix `DILUENT` or `NONE`, never `OXYGEN`), so `_resolveTankGasIndex`
  fell through to its "primary mix" guess and mislabeled the pure O2 tank
  with the diluent's composition (e.g. an 18% trimix diluent) instead of
  100% O2. That guess only makes sense for a tank actually breathed OC, so
  it's now skipped for a tank the device already tagged `DC_USAGE_OXYGEN`,
  and the resolver defaults that case to 100% O2 / 0% He instead.

## Test Plan

- [x] New unit test in `parsed_tank_resolver_test.dart`: an HP CCR oxygen
      tank with no matched gas mix now resolves to 100% O2 instead of the
      diluent's composition; a sibling diluent tank in the same dive is
      confirmed unaffected.
- [x] New widget test in `tank_editor_test.dart`: selecting `Oxygen Supply`
      in the role dropdown fills the O2/He fields with 100/0.
- [x] Full `dive_computer` and `dive_log` test suites pass (572 tests)
- [x] `flutter analyze` clean on changed files
- [x] `dart format` clean
- [x] Manually verified in a Windows build: switching a cylinder's role to
      `O₂ Supply` in the dive editor fills 100% O2 / 0% He


alpheios-one:
@ericgriffin All CI checks are green (Analyze, all 16 test shards, Android/iOS/macOS/Linux/Windows builds, CodeQL, Codecov). Could you take a look and test this for merge? Thank you very much for reviewing!

alpheios-one:
@ericgriffin Addressed the Copilot review finding (the unlinked-oxygen-tank guard now matches an explicit oxygen-tagged gas mix before falling back to gasless, so it no longer creates a duplicate phantom cylinder) and rebased onto current main to resolve the merge conflict. Also caught and fixed a related edge case during self-review: the role-switch auto-fill now feeds `_lastValidO2`/`_lastValidHe` too, so it plays correctly with the #1900 retype-fallback. All green again, ready for another look whenever you have time.

PR #1978, opened 2026-09-14
feat(import): keep the tissue loading the dive computer recorded (#1977)

### LINKED ISSUE #1977: Imported dives drop the tissue loading the dive computer recorded
My dive computer records its own tissue loading, but after importing the dive Submersion shows only its own recomputed values. The GF99 curve, the end-of-dive gradient factor and the compartment loading on the tissue card never match what the computer displayed during the dive, and there is no way to see what the computer actually reported.

This affects every import path that carries such data:

- Garmin Descent FIT files record an N2 tissue load percentage on every sample and a start and end N2 load in the dive summary.
- Shearwater Cloud database exports record GF99 on every sample plus the start surface GF and end GF99 of the dive.
- UDDF files written by Shearwater Cloud and Subsurface carry a `gradientfactor` value on every waypoint.
- Suunto logs (SML data, also delivered through Suunto Cloud) carry the per-compartment nitrogen and helium tensions at the start and end of the dive, the RGBM factors, CNS and OTU.
- Ratio iX3M XML exports carry 16 compartment loading percentages.

All of these are silently dropped on import.

Import the tissue data the computer recorded and show it wherever the app has a matching value. Where the computer reported a value, that value should be the one displayed; only where it did not should the app fall back to its own Buhlmann calculation from the profile. Concretely:

- Per-sample GF99 and N2 load stored with the profile.
- A dive-level start and end tissue snapshot (compartment tensions or loading, aggregate N2 load, GF99, surface GF, CNS, OTU, algorithm name).
- The profile GF99 curve drawn from the computer's values when they exist.
- A "dive computer" section on the tissue card showing the recorded start and end state, with the calculated value as a secondary line where both exist.
- UDDF export writing the recorded GF99 back so a backup round trip keeps it.

Keeping the recompute only. It cannot reproduce the computer's numbers because the computer's algorithm (Suunto RGBM, VPM-B, Garmin's implementation) and its residual state from earlier dives are unknown to the app, so the two will always disagree and the diver has no way to check what the computer said.

**Additional context**
No format carries per-sample per-compartment tensions except Ratio, so the app's 16-compartment recompute remains the only source for the tissue curves and the 3D tissue scene. Imported data replaces the calculated values only for the metrics the computer actually reported.

dotanalon:
## Related Issue

Closes #1977

## Summary

Dive computers export their own tissue state and the app dropped all of it, showing only a Buhlmann recompute that never matched the numbers the diver saw underwater. This imports what each supported format actually records and shows the computer's value wherever one exists, falling back to the app's calculation only where the computer reported nothing.

What each format carries, verified against real files:

| Source | Per sample | Dive level |
|---|---|---|
| Garmin FIT | `n2_load` percent | start and end N2 load, CNS |
| Shearwater Cloud DB | `gf99` | start surface GF, end GF99, deco model, CNS |
| UDDF (Shearwater Cloud, Subsurface) | `gradientfactor` (GF99) | none |
| Suunto SML via Suunto Cloud | none | per-compartment N2 and He tensions at start and end (9 or 15 RGBM compartments), RGBM factors, CNS, OTU, algorithm |
| Ratio iX3M XML | 16 compartment percentages (first and last sample stored) | start and end compartment loading, CNS |

No format carries per-sample per-compartment tensions except Ratio, so the 16-compartment recompute remains the source of the tissue curves and the 3D scene.

## Changes

- Profile series codec v2 adds `gf99` and `n2_load` per-sample fields. A v1 blob still decodes with them null; an older peer stores a v2 blob and skips it until updated (existing forward-version handling), so the sync compatibility floor is unchanged.
- New `dives.computer_tissue_json` column (schema 222) holding a `ComputerTissueSnapshot` with start and end `ComputerTissueState` (compartment tensions in bar or loading percent, aggregate N2 load, GF99, surface GF, CNS, OTU, RGBM factors, algorithm). Tolerant JSON decode so a corrupt column cannot break dive loading.
- Garmin FIT: `FitSample.n2Load`, `FitSummary.startN2/endN2`, snapshot built in `fitTissueSnapshot`; wired through both the file import parser and the `ImportedDive` path.
- Shearwater Cloud DB: reads `dive_log_records.gf99` and `dive_logs` header values when those optional tables exist, infers the `currentTime` unit per dive, aligns GF99 onto the libdivecomputer samples by nearest time (`ShearwaterGf99Aligner`), spells out the integer `decoModel` code, and prefers `calculated_values_from_samples.EndGF99` over a zero placeholder in `dive_details`.
- UDDF: per-waypoint `<gradientfactor>` parsed in both readers via `parseUddfGradientFactorPercent` (integers as percent, fractions scaled); both UDDF writers emit it back.
- Suunto: normalizer hoists `StartTissue`, `EndTissue` and `Algorithm` from the cloud `DiveHeader`; `parseSuuntoTissue` accepts all three array encodings and converts Pascal to bar; the snapshot travels on `DownloadedDive` through the import service into the repository, where a re-import without a snapshot never clears a stored one.
- Ratio iX3M: `tissueGroupNPercent` read from the first and last sample into start and end states.
- Analysis: `overlayComputerDecoData` gains `gf99Source` (computer-first) and replaces the GF99 curve with the computer's samples when present; `ProfileAnalysis.n2LoadCurve` exposed; `MetricSourceInfo.gf99Actual` reports what was used.
- UI: `ComputerTissueSection` on the tissue card, shown only when a snapshot exists: algorithm, compartment bars or tensions in the diver's pressure unit, start versus end table, calculated value as a secondary line where both exist. New `diveLog_computerTissue_*` strings, l10n regenerated.
- Docs: import gap tracker row added. All 11 locales carry the new strings; non-English ones are machine-style translations that a native speaker may want to review.

Not in this PR: storing Ratio's per-sample compartment series (needs 16 new codec channels), a chart for the N2 load curve, a GF99 source badge in the legend, and a standalone Suunto SML file parser (only the cloud path is live).

## Test Plan

- [x] `flutter test` passes (full suite run locally; the only remaining failures are the known local-timezone tests, which pass under `TZ=UTC`, and the Linux ffmpeg smoke test's output-size assertion, both unrelated to this change)
- [ ] Manual testing on: not manually tested

Fixture-backed tests cover the real Garmin Descent trimix file (start 0, end 86, peak 191 N2 load), the Shearwater Cloud CCR export (end GF99 62), the Perdix 2 and Petrel 3 UDDF exports (GF99 on 195 of 385 samples), and the Ratio iX3M export (16 compartments). The Shearwater fixture's sample table is empty, so per-sample GF99 alignment is covered by synthetic SQLite databases only. Suunto tissue parsing is covered by unit tests built from real EON Core and HelO2 header values; no fixture in the repo carries the tissue block.

## Screenshots

Not captured; the new tissue card section renders only for dives that carry an imported snapshot, and no such dive exists in the test fixtures that run in a widget harness.


dotanalon:
@ericgriffin can we merge this?

Issue #1427, opened 2026-08-31
Water type statistics


mboden:
**Describe the bug**
It seems like the statistics only looks at the water type directly assigned to dives and doesn't include dives with no water type set but where the dive site has a water type.

<img width="1909" height="314" alt="Image" src="https://github.com/user-attachments/assets/5f0ecb3d-5e46-4133-8ff1-b3a2306ef12a" />

**Expected behavior**
That the statistics should include dives where the dive site has a water type. Also it would be nice if the dive inherited the water type from the dive site.

**Platform details:**
- OS: Windows
- Version 1.7.6.7161

Issue #793, opened 2026-07-31
Recording the characteristics of the water body at a dive site / during a dive


alpheios-one:
The water type is very closely linked to the dive site – this is simply the nature of the system.
If a dive is recorded for a dive site that has not been assigned a water type, the logbook automatically enters the water type ‘not specified’ for that dive.
If the water type is subsequently added under the dive site, the water type is not updated for the dives associated with that site.
It would be better if, when the water type is defined for a dive site, the dives associated with that site were also updated.

This should help many users who are migrating an existing logbook to Subversion and are continuously populating the data.

It is possible that other fields may also be affected by this mechanism.

PR #1979, opened 2026-09-14
fix(export): fall back to the dive site's water type and entry method

alpheios-one:
## Summary

UDDF export, the PADI and Detailed PDF logbook templates, and the customizable
table view all read `Dive.waterType`/`Dive.entryMethod` directly. A dive
logged before its site had a water type set — or whose site gained one
later — still showed nothing there, even though the dive detail page and
statistics already resolve this correctly via
`Dive.effectiveWaterType`/`Dive.effectiveEntryMethod` (added in #1427).

This switches the four remaining raw readers to the same fallback getters,
so exports and the table view stay consistent with what the app already
shows on screen.

## Changes

- `lib/core/constants/dive_field_extractor.dart` — table view / customizable
  columns
- `lib/core/services/pdf_templates/pdf_template_detailed.dart`
- `lib/core/services/pdf_templates/pdf_template_padi.dart`
- `lib/core/services/export/uddf/uddf_export_service.dart` and
  `uddf_export_builders.dart` (bulk export and full-export paths)

No schema change, no migration — `dive.waterType`/`dive.entryMethod` keep
storing their own value; only the read side now falls back to the site's
value when the dive has none.

## Testing

Added fallback-coverage tests for all four call sites (site has a value,
dive doesn't → site's value is used; dive has its own value → it wins over
the site's):

- `test/core/constants/dive_field_extractor_test.dart`
- `test/core/services/pdf_templates/pdf_detailed_content_test.dart`
- `test/core/services/pdf_templates/pdf_padi_water_type_fallback_test.dart` (new)
- `test/core/services/export/uddf/uddf_export_builders_test.dart`

## Out of scope

- `visibility`, `altitude`, `bodyOfWater` have no `effective*` fallback and
  are not touched here.
- No write-back into the `dives.waterType`/`dives.entryMethod` columns
  themselves.

Refs #793


alpheios-one:
Hi @ericgriffin — this one's green across the board now (all CI checks passing, full coverage) and ready whenever you have time to take a look or give it a try. Thanks so much for picking up #793!

PR #2042, opened 2026-09-17
fix(dive-list): stop the detailed card stat row overflowing at narrow widths

### LINKED ISSUE #2041: Detailed dive card stat row overflows at narrow master-pane widths
## Bug

On the detailed dive card in the dive list, the stat row (the two stat slots plus the right-aligned dive-type badges) overflows horizontally when the master pane is narrow, painting the yellow/black overflow stripe.

`_kMasterPaneMinWidth` in `lib/shared/widgets/master_detail/master_detail_scaffold.dart` is 280, so a diver can drag the dive list pane to exactly a width where this happens.

## Measurements (main, dive with three dive types, tile in a fixed-width box)

| Width | Overflow |
| --- | --- |
| 240px | 52px |
| 260px | 32px |
| 280px | 12px |
| 300px | 21px |
| 320px | 1px |
| 340px and up | none |

With large but realistic stat values (131.4m, 147 min) and four dive types, it overflows up to about 340px. A joined dive-type list in a stat slot overflows by 250-330px at every width from 280px to 360px.

## Root cause

In `DiveListTile.build` (`lib/features/dive_log/presentation/pages/dive_list_page.dart`) the stat row's first child is a `Row(mainAxisSize: MainAxisSize.min)` holding two rigid `Text` stats with no `maxLines`/`overflow`, so nothing can shrink once the stats alone exceed the line.

There is a second overflow inside `DiveTypeBadgeRow`: it always renders at least its collapsed "+N" badge, so when the stats leave its `Expanded` narrower than that badge, the badge row itself overflows (seen at 320-340px).

Wrapping the stats in `Flexible` does not work: next to the badges' `Expanded`, two flex-1 children split the width 50/50 regardless of content, and the badge cluster overflows by 54-110px instead.

## Expected

No overflow at any master-pane width. The badges collapse to "+N" first; only then do the stats give way, ellipsizing.

## Note

PR #2013 moves the OC/CCR mode badge onto this row, which shifts the overflow-free threshold from about 340px to 353px. That PR deliberately leaves this bug alone.

ericgriffin:
## Related Issue

Closes #2041

## Summary

The detailed dive card's stat row overflowed horizontally when the dive list master pane was narrow, including at its 280px minimum. The two stats were rigid `Text` in a `Row(mainAxisSize: min)`, so nothing could shrink. At 320-340px the stats also squeezed the badges' `Expanded` below the width of their collapsed "+N" badge, so `DiveTypeBadgeRow` overflowed as well.

`Flexible` on the stats is not a fix: next to the badges' `Expanded` it splits the line 50/50 regardless of content. Flex allocates in a single pass and cannot express "collapse the badges first, then shrink the stats", so the line is now laid out from measurements.

## Changes

- New `DiveCardStatRow` (`lib/features/dive_log/presentation/widgets/dive_card_stat_row.dart`) owns the stat line. Inside a `LayoutBuilder` it:
  1. reserves the badge cluster's minimum (its collapsed "+N" badge plus the gap),
  2. gives the stats their natural `TextPainter`-measured width from the remainder,
  3. when that is too small, splits the budget by fair share (`fairShareWidths`): a stat narrower than an equal share keeps its full width and the other stat takes the rest, so a short depth never loses characters to a long neighbor. Stats ellipsize on one line, and an icon is dropped only if its slot is narrower than the icon itself.
  The badges' `Expanded` then gets whatever the stats leave, expanding back out of "+N" as room allows.
- `DiveTypeBadgeRow.minWidthOf` exposes the collapsed width using the same measurement the row uses internally. That measurement now honors `MediaQuery.textScalerOf`, so the reservation matches the rendered badge under larger text.
- `DiveListTile` builds a `DiveCardStat` per slot (`_buildStat`, replacing `_buildStatWidget`) and hands them to `DiveCardStatRow`.

### Interaction with #2013

#2013 has merged and main is merged in here. Its layout is kept: the mode badge is the right-most badge, and inside the badge cell the type badges collapse to "+N" while the mode badge keeps its width. `DiveCardStatRow` now owns that cell and reserves the mode badge's full width through the new `DiveModeBadge.widthOf` (text-scale aware, like the type badge measurement). The mode badge is always shown, so the reservation applies even to a dive with no types.

## Test Plan

- [x] Regression test written first and watched fail on main: `dive_list_tile_stat_row_overflow_test.dart` renders `DiveListTile` at 280-360px (10px steps) with four dive types and asserts `tester.takeException()` is null, for the default stats and for a stat slot holding the joined dive-type list. On main, 14 of 18 cases failed (overflows of 6-37px for default stats, 254-334px for the long stat).
- [x] `dive_card_stat_row_test.dart`: fair-share allocation cases, a no-overflow sweep from 120px to 400px, a short stat is not truncated beside a long one, and nothing truncates on a wide line.
- [x] Tile cases for an icon-less stat field ("<label>: <value>") at 280-360px.
- [x] Badge reservation holds under a 2x text scale (120-400px sweep); removing the text scaler from the measurement turns it red.
- [x] A stat keeps its icon from a 14px slot upward (the gap shrinks first), swept at 1px steps.
- [x] Mutation checks: removing the badge reservation turns 18 cases red; removing only the mode badge reservation (after merging #2013) turns 7 red.
- [x] `flutter test test/features/dive_log test/features/dashboard test/architecture` passes (4804 tests after merging main)
- [x] `flutter analyze` passes (whole project)
- [ ] Manual testing on: not run; covered by widget tests at fixed widths


ericgriffin:
Both suppressed comments from [this review](https://github.com/submersion-app/submersion/pull/2042#pullrequestreview-5230681537) are addressed in bdf4c1c:

- **`dive_card_stat_row.dart`: icon dropped at 14-18px.** The icon is now dropped only when the slot is narrower than the icon itself (14px). Between 14px and 18px the gap shrinks to what is left and the text gets zero width. `a slot keeps its icon while the icon itself fits` sweeps the slot width in 1px steps, reads each slot's rendered width, asserts no overflow and that both icons show from 14px up, and checks that the sweep actually covered the 14-18px range. It fails against the old threshold.
- **`dive_type_badge_row.dart`: scaled measurement untested.** `the badge reservation follows the text scale` runs the row under `MediaQuery` with `TextScaler.linear(2)` across 200-400px with type badges, the mode badge and a long stat, and asserts `takeException()` is null, so neither the badge cell nor the stat row may overflow. Removing `textScaler` from the badge measurement turns it red. After merging #2013, the new `DiveModeBadge.widthOf` measurement is scale-aware too and runs through the same test.

PR #2077, opened 2026-09-18
fix(media): start every photo picker session with no gallery selection

### LINKED ISSUE #1996: Media selection : not reset beetween entities
Hi,

It is something I noticed with dive sites (not tested with dives)

- Dive site A : I select some medias, click on Finish
- I go on Dive site B, I want to add some medias, the medias I selected for Dive Site A are still selected

The media should not be preselected

behaviour seen on Android ; not tested on Windows

ericgriffin:
## Related Issue

Closes #1996

## Summary

Photos ticked in the gallery picker while attaching media to one dive site came back pre-selected when the picker was opened for the next site. The selection lives in `photoPickerNotifierProvider`, which is app-scoped (not autoDispose), and nothing reset it when a new picker session opened.

The leak was not site-specific: dives, species photos and the library importer all open the same `PhotoPickerPage`, so they all inherited the previous session's ticks. It was also not only cosmetic: if the stale asset IDs fell inside the new owner's date window, tapping Done would have attached the previous owner's photos as well.

## Changes

- `PhotoPickerPage` already clears the Files tab's leftover staging when a session opens (`_clearStaleStaging`). It now clears the gallery selection in the same hook.
- The clear runs in the `initState` microtask, which is queued before the first `await` in the permission check and asset load, so the desktop path's auto-select of freshly picked files is never undone by it.
- New widget test, mirroring the existing Files-tab staging test: seed a picker notifier with two picks from "site A", open the page for "site B", and assert the selection is empty. It fails without the fix and passes with it.

## Test Plan

- [x] `flutter test test/features/media/presentation/pages test/features/media/presentation/helpers` passes (168 tests)
- [x] New regression test verified red before the fix, green after
- [x] `flutter analyze lib/features/media test/features/media` passes
- [x] `dart format .` makes no changes
- [ ] Manual testing on: Android (reporter's platform), select media on site A, tap Done, open the picker on site B and confirm nothing is pre-selected


ericgriffin:
Closing as a duplicate of #2082, which fixes the same issue (#1996) and also clears the URL tab's leftover draft, with broader test coverage.

Issue #2094, opened 2026-09-18
Media sync slice 2: Media health report, media log category and named origin device


ericgriffin:
Slice 2 of #2090.

Spec 4.2 to 4.4; plan Tasks 5 to 10. `LogCategory.media` with an info file floor, media breadcrumbs routed to it, a `PeerDeviceNameStore` fed from peer manifests during pull and adopt, the origin device label on placeholders and the info panel, and `MediaHealthReporter` surfaced from the Media info panel (copy), the Media Storage page (export) and the debug log export (bundled).

Issue #2090, opened 2026-09-18
Media sync program: make photo and video sync reliable end to end


ericgriffin:
## Goal

A photo or video linked on one device shows, with bytes, on every other device, on every platform pair; media transfers never wait silently; deletes propagate. This issue tracks the program end to end.

Design: `docs/superpowers/specs/2026-09-18-media-sync-program-design.md`. Phase 0 plan: `docs/superpowers/plans/2026-09-18-media-sync-phase0-harness-and-diagnostics.md` (both on branch `ericgriffin/media-sync-program`).

## Why one program

Media sync is three systems the user experiences as one feature: the `media` row travelling through the changeset engine, the peer re-finding the bytes in its own library or on disk, and the media store copies drained by the transfer queue. Nearly every report is a failure at a handoff between two of them, so single-issue fixes keep landing and reports keep coming.

Open issues folded in: #425, #1937, #1625, #2018, #1954. Closed by workaround, same class: #1968, #1661. External blocker with no code slice: #1738 (Dropbox developer console).

Verified on main `f3094a71228`: the media tables merge as blind upserts (the HLC stale-copy guard never runs for them); a locally pending row makes the engine skip the peer's whole update and the cursor still advances; `stampVerification` marks a row pending on every verify outcome; the gallery resolver on iOS, macOS and Android never answers "from other device"; Android limited photo access is treated as full access; no test anywhere simulates two devices exchanging a media row.

## Slices

Each slice is one PR whose description says `Part of #<this issue>` plus `Closes #N` for any issue it resolves. Dependencies run top to bottom.

- [x] 1. #2091 Two-device harness with seed scenarios S1 to S10 (skipped red on main)
- [x] 2. #2094 Media health report, `LogCategory.media`, peer device names, named origin device
- [x] 3. #2097 Engine merge rule: merge and keep pending, per-entity fact columns, media tables join the clocked guard (S1, S3)
- [x] 4. #2100 Quiet verification, `mediaStores` parent refs, child tombstones (S2)
- [x] 5. #2103 Google Drive error message and adapter audit (closes #2018)
- [x] 6. #2108 Diver delete media cascade (S9, closes #1954)
- [x] 7. #2113 Origin-aware gallery verdicts on iOS, macOS and Android (S5)
- [x] 8. #2116 PhotoKit cloud identifier: schema, stamp, resolve, backfill, cache invalidation (S6, closes #1937)
[...]

PR #2150, opened 2026-09-19
feat(media): media log category, peer device names and the media health report

ericgriffin:
Slice 2 of the media sync program: the diagnosis path. Stacked on #2140 (slice 1); the base moves to main once that merges, and CI runs then.

## What this adds

- **`LogCategory.media`** with a per-category file floor at info, so resolver and transfer worker breadcrumbs reach the log a user exports even with verbose logging off. A category floor can only admit more than the global floor, never less. The viewer's filter chips, colours and default set know the new value; all 11 catalogs name it.
- **Media loggers pinned to the category.** `LoggerService` gains a per-instance default category (still `const`), and the asset resolution service, both file and store resolvers, the upload pipeline and the transfer worker use it.
- **`PeerDeviceNameStore`.** The changeset reader records each peer's manifest name as it parses it on pull, and the adopt path does the same; a change stream and a live map provider let labels already on screen update when a sync learns a name.
- **Named origin device.** Both resolvers that answer "from other device" take a label lookup, so the placeholder reads "From Eric's MacBook"; the info panel's Linked on row shows the peer's name and updates in place.
- **`MediaHealthReporter`.** Per row: the synced facts and clock, the pending mark, the local asset cache verdict with its next retry, a fresh resolver verdict, the store object's presence in whichever namespace the stamps name, and the queue entry; every report carries the device and store header. The store is reached through a new side-effect-free `attachedMediaObjectStoreProvider`, extracted from the runtime provider, so a report never starts a drain or a sweep.
- **Three entry points.** Copy diagnostics on the Media info panel (one row, clipboard), Export media report on the Media Storage page (whole library, share or save, with a note that the report lists file paths and device names), and the debug log share, which bundles the library report as a second file and still ships the log if the report cannot be built.

## Verification

- `flutter analyze` on the whole project: no issues.
- `flutter test test/features/media test/features/media_store test/features/settings test/core/services test/core/models test/architecture`: 7333 passed, 20 skipped (the 10 seed scenarios plus pre-existing skips), 0 failed.
- The reporter is tested end to end on the two-device harness from slice 1; the peer name flow is tested with two real sync services over one fake cloud.

Design: `docs/superpowers/specs/2026-09-18-media-sync-program-design.md`; plan Tasks 5 to 10 in `docs/superpowers/plans/2026-09-18-media-sync-phase0-harness-and-diagnostics.md` (both in #2137).

Part of #2090
Refs #2094


lib/features/media/data/services/media_health_reporter.dart
```diff
@@ -0,0 +1,252 @@
...
+              0,
+              _backoffLadder.length - 1,
+            )],
+          );
+    final hlc = await _mediaRepository.getSyncHlc(item.id);
+
+    String verdict;
+    try {
+      final data = await _registry.resolverFor(item.sourceType).resolve(item);
+      verdict = data is UnavailableData ? data.kind.name : 'available';
```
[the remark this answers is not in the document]
ericgriffin: Fixed in 4fbb4c55b7f: the report now takes each row's verdict from verify(), the check API, instead of resolve(), the bytes API, and a cloud-backed row skips the registry entirely because its resolver reaches mediaStoreResolverProvider, which watches the runtime. The store probe is the verdict for those rows. A regression test asserts a mediaStore-source row reports without touching the registry.

lib/features/media/data/services/media_health_reporter.dart
```diff
@@ -0,0 +1,252 @@
...
+    final hash = item.contentHash;
+    if (hash == null) return (exists: null, tier: null);
+    final store = await _store();
+    if (store == null) return (exists: null, tier: null);
+    final ext = StoreKeys.extensionFor(item.originalFilename);
+    final stamped = <(String, String)>[
+      if (item.remoteUploadedAt != null)
+        ('original', StoreKeys.objectKey(hash, extension: ext)),
+      if (item.remoteCompressedUploadedAt != null)
+        ('rendition', StoreKeys.renditionKey(hash, ext: ext)),
```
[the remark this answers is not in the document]
ericgriffin: Fixed in 4fbb4c55b7f: the rendition probe uses the compressed extension the pipeline writes (mp4 for a video, jpg for an image), not the original filename's, with a test that seeds an mp4 rendition for a .mov row and expects the probe to find it.

lib/features/media_store/presentation/providers/media_store_providers.dart
```diff
@@ -391,20 +391,18 @@ final mediaStoreServiceProvider = Provider<MediaStoreService>(
...
+/// sweep, so diagnostics can read the store without changing anything.
+/// [mediaStoreRuntimeProvider] builds on top of this.
+// no-tick: builds a store ADAPTER from attach state, not a cached query
+// result. The one repository read (the connected account behind the
+// attachment) changes only on connect or disconnect, and both invalidate
+// this provider together with the runtime; a tick would rebuild the adapter
+// under a drain in flight.
+final FutureProvider<MediaObjectStore?> attachedMediaObjectStoreProvider =
+    FutureProvider<MediaObjectStore?>((ref) async {
       final attachState = ref.watch(mediaStoreAttachStateProvider);
```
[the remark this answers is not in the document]
ericgriffin: Fixed in 4fbb4c55b7f: invalidateMediaStoreAttachment now invalidates attachedMediaObjectStoreProvider first, since the runtime is built from it, and its doc lists all three cached providers. The new test was mutation-checked: with the line removed it fails, with it present it passes.

lib/core/services/sync/peer_device_name_store.dart
```diff
@@ -0,0 +1,58 @@
...
+  String? nameFor(String deviceId) => all()[deviceId];
+
+  /// Remembers [name] for [deviceId]. An empty name and an unchanged name
+  /// are no-ops, so a sync that sees the same manifests emits nothing.
+  Future<void> record(String deviceId, String name) async {
+    if (name.isEmpty) return;
+    final current = all();
+    if (current[deviceId] == name) return;
+    final next = {...current, deviceId: name};
+    await _prefs.setString(prefsKey, jsonEncode(next));
```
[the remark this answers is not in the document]
ericgriffin: Fixed in 0e9e0f7. `record` now takes a nullable name and removes the entry when it is null or empty, and both recording sites (the pull in `ChangesetReader` and the adopt scan in `SyncService`) pass whatever the manifest carried rather than skipping the call. They still skip only when the manifest could not be read at all, so a download or parse failure is not mistaken for a cleared name. Covered by a store-level test for the forget path and a two-device pull test where A publishes a name, B learns it, A republishes without one and B forgets it; the pull test goes red if either call site is put back behind the non-empty guard.

lib/features/media/data/services/media_health_report.dart
```diff
@@ -0,0 +1,241 @@
...
+  final String? queueState;
+  final int? queueAttempts;
+  final DateTime? queueNextAttemptAt;
+
+  /// True when the queue entry is parked until [queueNextAttemptAt].
+  final bool? queueWaiting;
+  final String? queueError;
+
+  /// The source pointer this row's type uses.
+  String? get pointer => filePath ?? localPath ?? platformAssetId;
```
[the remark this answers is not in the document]
ericgriffin: Fixed in 0e9e0f7. Rather than grow a second copy of the selection, `pointer` is now a field the reporter fills from `OriginFacts.from(item).pointer`, which is the exhaustive switch you pointed at, so network and manifest rows carry their URL, connector rows their remote asset id, store rows their content hash, and a bookmark-only local row its scoped reference. It is also emitted in the JSON, which previously had no locator at all for those types. A new test walks every source type and asserts both the field and the JSON value; it fails if the old three-column derivation is restored.

lib/features/media/presentation/providers/media_health_providers.dart
```diff
@@ -0,0 +1,62 @@
...
+    syncRepository: SyncRepository(),
+    assetCache: ref.watch(localAssetCacheRepositoryProvider),
+    queue: ref.watch(mediaTransferQueueRepositoryProvider),
+    registry: ref.watch(mediaSourceResolverRegistryProvider),
+    attachState: ref.watch(mediaStoreAttachStateProvider),
+    store: () => ref.read(attachedMediaObjectStoreProvider.future),
+    localDeviceId: () => SyncRepository().getDeviceId(),
+    localDeviceName: () async =>
+        (await SyncDeviceMetadata(SyncRepository()).resolve()).name,
+    deviceName: (id) => ref.read(peerDeviceNamesProvider).value?[id],
```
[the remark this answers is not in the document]
ericgriffin: Fixed in 0e9e0f7. The provider now uses `ref.read(peerDeviceNameStoreProvider).nameFor(id)`, matching the resolvers. Added a provider test that builds a report for a foreign row through the real provider graph and asserts both that the row is named and that `peerDeviceNamesProvider` was never started; it fails on the previous `.value?[id]` read.

lib/core/services/sync/sync_service.dart
```diff
@@ -4166,6 +4175,9 @@ class SyncService {
       } catch (_) {
         continue;
       }
+      // Recorded unconditionally: the manifest parsed, so a missing name is
+      // the peer's current state and clears any name it published before.
+      await _peerNames?.record(deviceId, manifest.deviceName);
```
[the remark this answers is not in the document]
ericgriffin: Fixed in ba171ad. The adopt-scan write now carries its own try/catch and logs a warning, so a preferences failure no longer aborts `_collectEpochBaseSources` and the library still adopts. The pull path needed the same treatment for a different reason: its write sat inside the per-peer `try`, so the same failure was recorded as a failed peer read and cost that peer its changesets for the cycle and left its cursor unmoved. Both sites are guarded now. Two regression tests, one per path, and each goes red when its own guard is removed.

lib/features/settings/presentation/providers/debug_log_providers.dart
```diff
@@ -184,10 +191,16 @@ Future<void> shareLogFile(
   final tempDir = await getTemporaryDirectory();
   final export = File('${tempDir.path}/$_exportFileName');
   await export.writeAsBytes(await buildLogExportBytes(file, header));
+  final files = [XFile(export.path, mimeType: 'text/plain')];
+  if (mediaReport != null) {
+    final report = File('${tempDir.path}/$_mediaReportFileName');
+    await report.writeAsString(mediaReport);
```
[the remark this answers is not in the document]
ericgriffin: Agreed, and fixed in e28b26e. Rather than redact at the one write site, every rendering of the report now goes through `redactSecrets`: both `toText` paths and the string values of both `toJson` maps. That covers all three exits, the log share, the Media Storage export and the info panel's clipboard copy, so a future consumer cannot reintroduce the gap. Queue errors are covered too, since they are an exception `toString()` and can carry a request URL or header. A test pins both directions: a signed URL, a `user:password@` host and a token inside a queue error are masked, while file paths, device names and content hashes pass through untouched. Verified by reverting each redaction point separately, which turns the corresponding assertions red.

lib/features/media/data/services/media_health_reporter.dart
```diff
@@ -0,0 +1,330 @@
...
+            ),
+          );
+    final cacheNextRetryAt =
+        (cache == null || cache.resolutionMethod != 'unresolved')
+        ? null
+        : DateTime.fromMillisecondsSinceEpoch(cache.resolvedAt).add(
+            _backoffLadder[cache.attemptCount.clamp(
+              0,
+              _backoffLadder.length - 1,
+            )],
```
ericgriffin, line 181: The comment above promises `verify()`, never `resolve()`, but `LocalFileResolver.verify` opens with `await resolve(item)` (`local_file_resolver.dart:464`), so for `localFile` rows the guarantee does not survive one call frame.

`resolve`'s fallback branches are byte reads: on macOS and iOS a row whose `localPath` is absent or unreadable goes to `readBookmarkBytes` (`local_file_resolver.dart:342`), and on Android to `readUriBytes` (`local_file_resolver.dart:315`). Both return `BytesData` built from the entire file. Sandboxed macOS imports normally resolve through bookmarks by design, so `forLibrary()` there reads every photo and video fully into memory, one row at a time, serially. On a few thousand rows the export takes minutes with the page's whole action set disabled, and a multi-GB video can take it out of memory.

Either add `localFile` to `_skipReason`, or give the resolver a stat-only verify path that does not route through `resolve`.

lib/features/media/data/services/media_health_reporter.dart
```diff
@@ -0,0 +1,330 @@
...
+    // report would read every local file and decode every gallery asset.
+    //
+    // And only for the source types whose verify() is a local read. The
+    // others are reported from stored state, which is not a lesser answer:
+    // each has a dedicated service that owns the live check, and the row
+    // already carries what that service last wrote.
+    final skip = _skipReason(item.sourceType);
+    String verdict;
+    if (skip != null) {
+      verdict = skip;
```
ericgriffin, line 195: `_localDeviceName()` runs once per row whose origin is this device, and the injected implementation rebuilds the answer every time: `mediaHealthReporterProvider` passes `() async => (await SyncDeviceMetadata(SyncRepository()).resolve()).name`, and `resolve()` does a database read (`getDeviceId`), an uncached MethodChannel round trip (`DeviceDisplayNameService.resolve`) and `PackageInfo.fromPlatform()`.

Most rows in a library were linked on this device, so a 2000-row report issues 2000 database reads and roughly 4000 platform-channel calls for a value that cannot change mid-report. `_report` already resolves it once for the header; resolving it once in `forItem`/`forLibrary` and passing it into `_row` removes the rest.

lib/features/media/data/services/media_health_reporter.dart
```diff
@@ -0,0 +1,330 @@
...
+    // The store resolver reaches mediaStoreResolverProvider, which watches
+    // the runtime, and building that starts a transfer drain and a verify
+    // sweep. The store probe is this row's verdict instead.
+    MediaSourceType.mediaStore => 'mediaStore',
+
+    // A gallery verify runs AssetResolutionService, which on a miss writes
+    // the local asset cache: the attempt count goes up and resolvedAt is
+    // reset to now, so the next automatic search moves from 24 hours out to
+    // 3 days and then a week. Producing a report would delay the recovery
+    // of the very photos it is describing. The cache block on this row is
```
ericgriffin, line 282: The originals probe keys off this row's own filename extension, but `StoreKeys.extensionFor` is documented as not injective over content (`objectKeyPrefix` exists precisely so callers can find the variants).

Device A holds the photo as `IMG_1.JPG` and uploads it to `smv1/objects/ab/<hash>.jpg`, then stamps `remoteUploadedAt`, which syncs. Device B holds identical bytes as `IMG_1.jpeg`, and its pipeline skips the upload because the stamp is already set (`media_upload_pipeline.dart:90`). Copy diagnostics on device B probes `<hash>.jpeg`, `head` returns null, and the report prints `remote_uploaded_at: <ts>` directly above `store_object: missing`. In a support thread that reads as store corruption when the object is there.

Listing `StoreKeys.objectKeyPrefix(hash)` rather than heading one exact key would answer correctly.

lib/features/media_store/presentation/pages/media_storage_page.dart
```diff
@@ -833,6 +860,23 @@ class _MediaStoragePageState extends ConsumerState<MediaStoragePage> {
                   padding: EdgeInsets.only(top: 8),
                   child: LinearProgressIndicator(),
                 ),
+              const SizedBox(height: 8),
+              FilledButton.tonal(
+                key: const Key('media-export-report'),
```
ericgriffin: This button sits inside the `if (connected)` block at line 743, so the whole-library report is only reachable once a media store is attached.

Nothing in the report needs one: `_probeStore` and the marker read both degrade to "not probed", and the rows that matter most for a photo-sync complaint (local file and gallery rows, synced by row rather than by bytes) are exactly the ones a user with no store has. Someone diagnosing "my photos are not on my iPad" opens this page and sees only the connect form. The debug log share still bundles the report, but that is a second and much less discoverable path.

Worth hoisting the button and its note out of the connected branch.

lib/features/settings/presentation/providers/sync_providers.dart
```diff
@@ -489,6 +490,34 @@ final cloudStorageProviderProvider = Provider<CloudStorageProvider?>((ref) {
...
+  return store;
+});
+
+/// The live name map: the store's contents now, then every change, so a
+/// label already on screen updates when a sync learns a name.
+final peerDeviceNamesProvider = StreamProvider<Map<String, String>>((ref) {
+  final store = ref.watch(peerDeviceNameStoreProvider);
+  // Subscribe first, then snapshot. Yielding the snapshot and subscribing
+  // afterwards drops a name recorded in between, and the store emits only
+  // when a name CHANGES, so a peer that keeps publishing the same name
```
ericgriffin, line 507: `yield` suspends the generator before `yield*` subscribes, and `_changes` is a plain broadcast controller with no replay, so a `record` that completes in that gap is dropped.

Concretely: a pull is already in flight when a tile first watches this provider. The generator emits the current (empty) map and suspends, `ChangesetReader`'s `peerNames.record` calls `_changes.add` with no listener attached yet, then `yield*` subscribes. The Linked on row keeps the generic label until some later sync happens to record a different name, which is the staleness the live stream was added to prevent.

Subscribing first and emitting `store.all()` after, or re-reading it once the subscription is live, closes the window.

ericgriffin (commented):
Correctness pass over the diagnosis path. Seven findings, inline.

The one I would fix before merge is the first: `MediaHealthReporter` documents that it never takes the bytes path, but `LocalFileResolver.verify` calls `resolve` internally, so on macOS and Android a whole-library report reads every local file's bytes.

Checked and deliberately not flagged: the per-category file floor in `LoggerService._persists` is correct (it only ever lowers the floor, and non-verbose is `warning`, so the media `info` floor genuinely admits more); the `attachedMediaObjectStoreProvider` extraction preserves the runtime's behaviour and the invalidation helper now covers all three caches; and `_collectEpochBaseSources` does record this device's own manifest name, against `PeerDeviceNameStore`'s documented contract, but every consumer guards on `origin != me` first, so there is no reachable failure today.

lib/features/media/presentation/widgets/media_info_panel.dart
```diff
@@ -619,6 +622,39 @@ class _CopyReferenceButton extends StatelessWidget {
...
+  const _CopyDiagnosticsButton({required this.item});
+
+  final MediaItem item;
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) => TextButton(
+    onPressed: () async {
+      final messenger = ScaffoldMessenger.of(context);
+      final l10n = context.l10n;
+      // Guarded like the other two entry points. Unhandled, a report that
```
ericgriffin, line 638: No try/catch and no in-flight guard here, unlike the comparable actions on the Media Storage page.

`forItem` starts with `_localDeviceId()` (`SyncRepository().getDeviceId()`) and `_syncRepository.getPendingRecords()`, neither of which is guarded inside the reporter, so a database failure throws into the framework: no snackbar, nothing on the clipboard, and a button that looks inert. Separately, `probeStore: true` issues up to three network HEADs, so on a slow link the button stays enabled and a second tap starts a second report.

lib/features/media/data/repositories/media_repository.dart
```diff
@@ -2107,6 +2107,19 @@ class MediaRepository {
   /// A display label (file name, else caption) for each of [ids] that has
   /// one. Ids with neither, or with no row, are absent.
   ///
+  /// The row's sync clock, or null for an unknown id or a row never marked
```
ericgriffin: This method landed between the first paragraph of `getDisplayLabels`' doc comment and the rest of it, so the two now carry each other's documentation.

`getSyncHlc` opens with "A display label (file name, else caption) for each of [ids] that has one. Ids with neither, or with no row, are absent.", which describes a different method, and `getDisplayLabels` is left with the "selectOnly projection, one query per chunk" note stripped of its subject.

lib/features/media/data/services/media_health_reporter.dart
ericgriffin: Withdrawing this one: it is already fixed at HEAD. `_row` dispatches through `DiagnosticProbe` and only falls back to `verify` for resolvers that offer no probe, so `LocalFileResolver` never reaches the bookmark byte read. `media_health_reporter_verify_test.dart` pins both halves.

The review was built against a stale copy of the diff that predated the last few commits on the branch, which is my error, not the code's.

lib/features/settings/presentation/providers/sync_providers.dart
ericgriffin: Withdrawing this one: already fixed at HEAD. The provider subscribes to `store.changes` before snapshotting `store.all()`, with a comment covering exactly this race. Same stale-diff mistake as the reporter comment above.

lib/features/media/presentation/widgets/media_info_panel.dart
ericgriffin: Withdrawing the error-handling half: already fixed at HEAD. The callback wraps the report and the clipboard write in a try/catch and surfaces `settings_diagnostics_copyFailed`. Same stale-diff mistake.

The double-tap observation still stands in principle (the button stays enabled across the store HEADs) but it is minor on its own, so I am not asking for a change.

lib/features/media/data/services/media_health_reporter.dart
ericgriffin: Fixed in 091e6407b10. `forItem` and `forLibrary` resolve the name once and carry it into `_row` and `_report`. Covered by `media_health_reporter_probe_test.dart`, which counts the lookups: a five-row library went from six to one.

lib/features/media/data/services/media_health_reporter.dart
ericgriffin: Fixed in 091e6407b10. Each namespace is listed by its key prefix now (`objectKeyPrefix` / `renditionKeyPrefix`, and the thumb key which has no variants), short-circuiting on the first element so a paged listing stops after one page. Three tests cover it: the cross-extension hit, a hash with nothing stored, and a longer hash that shares this one as a prefix.

lib/features/media_store/presentation/pages/media_storage_page.dart
ericgriffin: Fixed in 091e6407b10. The button and its note are hoisted out of the `if (connected)` branch. New page test drives the export with no `statusHint`, so the disconnected path is pinned.

lib/features/media/data/repositories/media_repository.dart
ericgriffin: Fixed in 091e6407b10. `getSyncHlc` moved below `getDisplayLabels`, and the label doc comment is whole again.

PR #2189, opened 2026-09-20
refactor(dive-log): remove the unreachable playback UI from dive detail

### LINKED ISSUE #2184: Dive detail page carries unreachable playback UI that still rebuilds the page during fullscreen playback
**Describe the bug**

The dive detail page carries a block of playback UI that cannot be reached.

The "Playback" button that switched it on was removed deliberately in
c4e70ae1814 (2026-02-26, "Remove the playback button on dive details page until
the functionality can be more thought-out"), but the block it used to reveal
was left in place behind `if (playbackState.isActive)`. Nothing else on the
page can set `isActive`: the only widget that could is `PlaybackControls`, and
its activation branch lives inside the very gate it would have to open. The
fullscreen profile page is now the sole producer of `isActive`, and it clears
the flag again on dispose, so the detail page is guaranteed to read false.

This is not just inert. The detail page calls
`ref.watch(playbackProvider(dive.id))` unconditionally at the top of `build`,
solely to evaluate that gate, and it stays mounted under the fullscreen route
(`MaterialPageRoute`, default `maintainState`). Measured: `playbackProvider`
notifies 31 times per wall-second at the default 30x replay speed, so playing a
dive in fullscreen rebuilds the entire detail page about 31 times a second
behind it, to render a block that can never appear.

**To Reproduce**

1. Open any dive's detail page. There is no way to start playback.
2. Open the fullscreen profile and press play.
3. The detail page rebuilds on every playback tick underneath, invisibly.

**Expected behavior**

The unreachable block, the widgets only it referenced, and the provider watch
that exists only to evaluate it are removed. Playback itself is unaffected: the
fullscreen profile has its own transport and is the only surface that offers
it.

**Additional context**

Stranded by the gate:

- `playback_controls.dart` (240 lines), referenced only from the dead gate
- `playback_stats_panel.dart` (345 lines), referenced only from the dead gate
- `CompactTissueSaturation`, reachable only through the dead gate
- `diveLog_playback_stepThrough`, referenced only from dead code
- `diveLog_detail_button_playback`, already orphaned in all 11 ARB files with
  zero Dart references since c4e70ae1814

Dead independently, and superseded by the live tissue UI in `tissue_heat_map`,
`tissue_area_chart`, `compact_tissue_loading_card`, the surface-interval
`tissue_recovery_chart` and the `dive_3d` tissue widgets:

- `tissue_saturation_panel.dart` (233 lines): `TissueSaturationPanel` has had
  no caller since it was introduced
- `tissue_saturation_chart.dart` (332 lines), used only by that panel

No tests cover any of it.

ericgriffin:
Closes #2184

## The dead code

The dive detail page carried a block of playback UI nothing could reach.

The "Playback" button that revealed it was removed in c4e70ae1814
(2026-02-26), *"until the functionality can be more thought-out"*. That commit
took only the 17-line button; the block it revealed stayed behind
`if (playbackState.isActive)`. Nothing on the page can set that flag. The one
widget that could is `PlaybackControls`, whose activation branch sits inside
the very gate it would have to open. The fullscreen profile page became the
sole producer of `isActive`, and it clears the flag again on dispose, so the
detail page is guaranteed to read false.

## It was not inert

The page called `ref.watch(playbackProvider(dive.id))` unconditionally at the
top of `build`, only to evaluate that gate, and it stays mounted under the
fullscreen route (`MaterialPageRoute`, default `maintainState`).

Measured, not assumed: the detail page is still mounted while fullscreen is up,
and `playbackProvider` notifies **31 times per wall-second** at the default 30x
replay. So playing a dive fullscreen rebuilt the entire detail page about 31
times a second behind it, to render something that could never appear.

## Removed

- The gated block, plus the `playbackProvider` and profile-analysis watches
  that only fed it. The analysis watch's own `activeDiveSourceProvider` read is
  still made a few lines below, so no subscription is lost.
- `playback_controls.dart` (240 lines) and `playback_stats_panel.dart` (345),
  reachable only through the gate.
- `tissue_saturation_panel.dart` (233) and `tissue_saturation_chart.dart`
  (332). `TissueSaturationPanel` never had a caller, not even in the commit
  that introduced it, and the chart was used only by that panel. Only
  `CompactTissueSaturation`, in the same file, was reachable, through the gate.
- 22 ARB keys across all 11 locales, found by baseline-diffing a repo-wide
  reference scan rather than trusting "unreferenced" (about 790 keys are
  already unreferenced for unrelated reasons). 19 were orphaned by this change;
  `diveLog_detail_button_playback` has been orphaned since c4e70ae1814, and
  `diveLog_playbackStats_pressure` and `diveLog_playback_speed_label` were dead
  strings inside the dead files.

The range-analysis button loses its `if (!playbackState.isActive)` wrapper,
which was always true. That is the only behaviour change visible to a user, and
it is a no-op.

## Not touched

Playback itself. The fullscreen profile keeps its own transport
(`profile_transport_controls`), which is the only surface that offers playback,
and `PlaybackNotifier` stays as its engine.

The live tissue surfaces: `tissue_heat_map`, `tissue_area_chart`,
`compact_tissue_loading_card`, the surface-interval `tissue_recovery_chart` and
the `dive_3d` tissue widgets. `diveLog_playbackStats_deco` and
`diveLog_tissue_title` stay, since the profile chart and heat map still use
them.

Historical plan docs that mention the deleted files are records and were left
alone. The one exception is the 3D flythrough spec, which named
`playback_stats_panel` as a component to reuse; that line now points at this
removal, since the file is recoverable from history and the HUD may want
different content against a 3D scene anyway.

## Tests

No tests covered any of the removed code. Full suite green: 30471 passed, 21
skipped. `flutter analyze` clean, `dart format` clean, l10n regenerated with
`flutter gen-l10n`.


ericgriffin:
Follow-up filed as #2231 for the chart host subscription raised in review: `DiveProfileChartHost` still watches playback to feed `playbackTimestamp`, and both of its consumers sit under the fullscreen route, so the cursor it draws is never visible while it rebuilds per tick. Left out of this PR because it is live wiring for a real chart feature rather than dead UI, and removing it cascades into the chart's own cursor code and its tests.

